PR #8, opened 2021-12-21
GWAS_QC done, master_control overhaul

hsun3163:
I have revamped the eqtl_data_analysis(originally master_control) notebook so that it shall be complex enough to perform the code generation function but hopefully not too complicated.

Please check if the data_preprocessing and qtl section of the updated eqtl_data_analysis notebook looks like what you have in mind. 

If yes, I will use this as the backbone to drive my downstream update.


gaow:
> so that it shall be complex enough to perform the code generation function but hopefully not too complicated.

It does not matter if it "looks" complicated or not, as long as it generates reasonable code for analysis. Currently I still see obvious issues with the command generation (eg hard coded `/home/hs3163/GIT/ADSPFG-xQTL/code/csg.yml`) but I assume that we need to finalize each module before we can edit these codes to generate for the new modules? 

> Please check if the data_preprocessing and qtl section of the updated eqtl_data_analysis notebook looks like what you have in mind.

`data_preprocessing` looks reasonable (minus normalization and residual analysis?); `qtl` section it is not clear what if we want to perform both cis and trans analysis.

PR #9, opened 2021-12-21
Update PEER

Gou-29:
Hi professor: this is the updated version of PEER factor analysis with all detailed documentations and minimum working examples using sample data from PEER. A small bug is listed.  I will wrap up the docker image for PEER and APEX tmr. Hope it can be finished by Thursday.


gaow:
@Gou-29 terrific! Do you think you can base your docker image on an existing docker config I have? eg check it here:

https://github.com/gaow/misc/blob/master/docker/base-notebook.dockerfile

this is avaiable from dockerhub as `gaow/base-notebook`.

gaow:
@Gou-29 one suggetsion is don't include numbers in section titles. Just use the proper level of `#` that's enough and is better Otherwise it looks good! I like it that you do show results generated. Do you think you can explain in a few words the take home information from those plots?

Gou-29:
I will try it then. I think PEER is very special in its dependent. It cannot work for a higher version of R as well as a lower version of R... I will work on it then to see what will happen. Also, I failed to download APEX so I will try to wrap APEX in the same docker file.

gaow:
Posted here: https://cumc.github.io/xqtl-pipeline/pipeline/data_preprocessing/covariate/PEER_factor.html 

> I think PEER is very special in its dependent. It cannot work for a higher version of R as well as a lower version of R...

In that case feel free to create its own minimal docker image!

gaow:
>  I failed to download APEX so I will try to wrap APEX in the same docker file.

I think it is best to make seperate images for different notebook modules. Then it will make more sense that each image is as small as possible. For APEX, do you think we can use a basic, < 100Mb Debian Linux image to execute?

PR #24, opened 2021-12-23
Docker file for PEER image

Gou-29:
Attached is the initial version of PEER Docker image. As stated before, this is the most strange one. Compared to the reference:
* Only included R dependence
* Failed to include x11() device, so the diagnosis plot may not be generated
* Size of this image reduced significantly (from around 3G -> less than 1G and can be further reduced)
* Deployed at:  https://hub.docker.com/r/gouwh/r-peer

### DIFF
--- a/pipeline/data_preprocessing/covariate/Dockerfile
+++ b/pipeline/data_preprocessing/covariate/Dockerfile
@@ -1,0 +1,117 @@
+# Base image
+FROM ubuntu:18.04
+
+# Maintainer and author
+LABEL maintainer="Wenhao Gou<wg2364@cumc.columbia.edu>"
+
+#### Build code for installing R ####
+# For SOS based xQTL workwlow implementation. With just set an environment for PEER-R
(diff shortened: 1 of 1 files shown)


gaow:
Thank you @Gou-29 . Could you elaborate this : Failed to include x11() device, so the diagnosis plot may not be generated ?

Gou-29:
@gaow : In R, we need x11 device to  call functions like png() / dev.off() that are interactive with the graphical device. For example, in MacOS, R requires a software called XQuartz to render all the pictures. It is required for even rendering rmd files. In Linux, a suspect that all required dependence can be installed by `apt-get build-dep -y r-base`. But when including this, the docker file will be huge (i.e. it will also install latex, nodejs etc.) . I tried to install some of them but end up failed.

gaow:
@Gou-29 I remember in Linux it's the cairo package? Perhaps including that will work?

Gou-29:
@gaow  Yes, but still it requires x11. I believe what we can do is to use an r docker image to see what will happen then. I will go back to this next week (since other works need to be done)

PR #25, opened 2021-12-23
sos sumstat merging pipeline

changebio:
1. testing on MWE (flip, reverse, and flip+reverse cases)
2. adding documentation
3. making it into a SOS workflow.


changebio:
add template yaml in it

PR #27, opened 2021-12-24
Draft formula for regression Z out Y

Gou-29:
Hi professor, I wrote a draft version of regression Z out Y in `Data preprocessing --> Covariate data preprocessing --> Computing covariate residuals`, please check and comment so. After your review and correction, I will work on the code then.

Also, I noticed that there is also a file `Data preprocessing --> Phenotype--> Residual_Y `. Are we actually going to change this file?


gaow:
Thank you @Gou-29 . It looks good for now although the question remains whether we do it in X as well. Let's not do it at this point. The only thing to notice is taht we need the output of Y_res to be in the same format as the input. 

Regarding Residual_Y.ipynb -- please remove this file. (@hsun3163 we will be using `remove_covariate.ipynb` under `covariate` folder because it might involve changing X as well down the road).

PR #43, opened 2021-12-28
Update PEER + APEX and dockerfiles

Gou-29:


gaow:
@Gou-29 we have a `data` directory in the root of the repo. But I think I mentioned it before, for test data let's use a Google Drive folder and not share it publicly and not via github. Please request access to this link: https://drive.google.com/drive/u/0/folders/1ahIZGnmjcGwSd-BI91C9ayd_Ya8sB2ed and upload data there. Please remove them from github.

PR #66, opened 2022-01-04
fixed duplicated chr:pos by using an extra ID as index of each SNPs

changebio:
1, using unique ID to deal with duplicated SNPs in different gene
2, adding intersect parameter to output intersection or union SNPs, default is False.
3, changing —keep_ambigous to boolean type, default is False
4, updating yaml file (adding a ID: CHR,POS row). `ID: CHR,POS` can be `ID: CHR,POS,GENE` or any other combinations.
5, updating description in pipeline.


gaow:
@changebio looks good, thanks! But shall we make the default ID `chr_pos_a0_a1` I believe this is the  case for now anyways? Notice that this ID will use adjusted a0 and a1 after strand flip, and should be consistent with a0 and a1 in target set.

changebio:
> @changebio looks good, thanks! But shall we make the default ID `chr_pos_a0_a1` I believe this is the case for now anyways? Notice that this ID will use adjusted a0 and a1 after strand flip, and should be consistent with a0 and a1 in target set.

The SNP ID is still `chr_pos_a0_a1`, But the index of each row is the `ID`, which will be removed in output.

changebio:
ID is used to distinguish different SNPs before strand flip. In here, duplicated ID will be removed.

gaow:
I see. My proposal is that we make the ID both the final SNP ID and the ID to use before strand flip. Of course, before strand flip you'll ignore a0 and a1 if they are part of the ID. The reason to do it is that we want the final SNP ID to be customizable, eg, we want the output SNP to have ID that contains the gene information. this is important for makeing the VCF format for QTL applications.

hsun3163:
> @changebio looks good, thanks! But shall we make the default ID `chr_pos_a0_a1` I believe this is the case for now anyways? Notice that this ID will use adjusted a0 and a1 after strand flip, and should be consistent with a0 and a1 in target set.

at the moment default is chr:pos_ref_alt, shall I change other places to accommodate the chr_pos_ref_alt format?

changebio:
> > @changebio looks good, thanks! But shall we make the default ID `chr_pos_a0_a1` I believe this is the case for now anyways? Notice that this ID will use adjusted a0 and a1 after strand flip, and should be consistent with a0 and a1 in target set.
> 
> at the moment default is chr:pos_ref_alt, shall I change other places to accommodate the chr_pos_ref_alt format?

All you need to do is adding a ID row in yml, but the ID row shouldn't be `ID:CHR,POS,A0,A1`(The reason as following my reply to Gao). It can be `ID:CHR,POS` or `ID:CHR,POS,GENE`.

changebio:
> I see. My proposal is that we make the ID both the final SNP ID and the ID to use before strand flip. Of course, before strand flip you'll ignore a0 and a1 if they are part of the ID. The reason to do it is that we want the final SNP ID to be customizable, eg, we want the output SNP to have ID that contains the gene information. this is important for makeing the VCF format for QTL applications.

I thought about that. But SNPs' ID and the ID shouldn't be the same. Because we need to use the ID to find overlapping SNPs. If we use chr_pos_`a0_a1` as the ID, we will miss strand fliped or reversed SNPs.

But if you want to add something to SNPs' ID, this can be done by another way. such as `SNP:CHR,POS,A0,A1,GENE`.

gaow:
> But SNPs' ID and the ID shouldn't be the same. Because we need to use the ID to find overlapping SNPs. If we use chr_pos_a0_a1 as the ID, we will miss strand fliped or reversed SNPs.

Please note in my previous reply, " Of course, before strand flip you'll ignore a0 and a1 if they are part of the ID. ". So my solution should still work. 

> But if you want to add something to SNPs' ID, this can be done by another way. such as SNP:CHR,POS,A0,A1,GENE.

I dont want to introduce another field. That seems redundant.

PR #72, opened 2022-01-06
Dockerfile generation

hsun3163:
Dockerfile can be automatically generated from singularity recipe now.
This is done via the same commands:
`spython recipe singularityfile dockerfile`

However one caveat for this method is that, it cant be used directly on the singularity file that are generated from spython.
The stage:spython-base statement will produce some errors. Renaming the stage or deleting this line altogether can solve this issue.


gaow:
@hsun3163 am sorry but @changebio has updated the summary stats merger pipeline and there is a conflict between your version and his. Could you try to resolve  the conflict?

gaow:
the automatic conversion is done in SoS : https://github.com/cumc/xqtl-pipeline/tree/main/container/singularity 

the stage line is removed using `sed`.

PR #78, opened 2022-01-10
bug fix on polyfun

hsun3163:
put back the correct version of polyfun, with some additional bug fix.

remaining problem:
1. Too high mem requirement, need to generate mwe
2. for some reason munged_sumstat cannot take gz file, contrary to provided script and example on polyfun github

Amanda will be looking into these two


hsun3163:
It is verified that the issue of unable to read txt.gz sumstat is caused by the container. However, it is uncleared what exactly cause the issue. The next step of figuring this out is to look into the actual script to see what went wrong.

PR #87, opened 2022-01-20
fix and regenerate def for bioinfo+fix in readCount QC

hsun3163:
1. Now readcount QC is bug free
2. Add a optional step for normalization module to transform the output from readcount QC, and other bed-like tpm/genecount table to what the normalization module like.
3. The bioinfo.def shall be correct. The reported bug no longer exist, however, a new error occurs that I am unsure whether it is the problem of the def or the remote builder:
```
FATAL: Unable to push image to library: error uploading file to server: Post "https://library.sylabs.io/v1/imagefile/hs3163/remote-builds/rb-61e8d768257059d32226312b:sha256.f905b5c20b7ad970d3fb6471409aaaf8687a155826c653ef3884e27e39089361": unexpected EOF
```
The provided link contains following info:
```
{"error":{"code":404,"message":"NOT_FOUND: repository hs3163/remote-builds/rb-61e8d768257059d32226312b not found"}}
```


gaow:
@hsun3163 the updated `def` file works with `--fakeroot` build. Thank you! (The error you experience looks like a 404 error exclusive to the `--remote` build ...)

hsun3163:
sylab have fixed the eof problem. Built success

PR #92, opened 2022-01-22
rna_quant docker/singularity done

hsun3163:
Fix various things in both the dockerfile and the def that will either break the building or makes the built image lacking certain components.

Script shall be able to called via something like 
```
bash: expand = "$[ ]", stderr = f'{_output}.stderr', stdout = f'{_output}.stdout' , container = container
         finemapper.py \
        --geno $[_input[0]:n] \
        --sumstats $[_input[1]]  \
        --n $[sample_size] \
```

bulk_rna module is still WIP


pipeline/molecular_phenotypes/bulk_expression.ipynb
```diff
@@ -343,7 +343,7 @@
     "# Path to the software:\n",
     "parameter: soft_dir = path\n",
     "# Path to the reference adaptors:\n",
-    "parameter: adapter = str\n",
+    "parameter: adapter = \"ILLUMINACLIP:TruSeq3-PE.fa:2:30:10\"\n",
```
gaow, line 346: Is it typically to have a default adaptor, or should we always ask people to input an adaptor? It seems particularly specific. We make something default only if it is a reasonable choice. We dont make defaults in order to lighten the interface.

hsun3163: fastqc should be able to tell what adaptor is there. But this is also part of the wip that haven't implemented yet.

gaow: fastqc is an interactive tool -- sounds like users need to run that on their own to determine the adaptor? In that case adapter should be made a required parameter without default. And we need to document how to get that sequence. It's not clear how this is obtained from the biostars link ... Do you know what to do here?

hsun3163: > fastqc is an interactive tool -- sounds like users need to run that on their own to determine the adaptor? In that case adapter should be made a required parameter without default. And we need to document how to get that sequence. It's not clear how this is obtained from the biostars link ... Do you know what to do here?

one of the existed step involved running fastqc on the fq file to generates the html version of the report and a zip file containing the text description and images that are used in the report. If the adaptor information can be found by fastqc, it should be in there. But I am still figuring out what exactly to extract.

gaow: Thank you @hsun3163 . I hope you can figure it out. I'll merge the PR after you figure it out and document it clear enough for users to locate and use it under your instructions. Let me know if you have an issue with it (we can use expertise from our other colleagues)

hsun3163: OK it seems the Trimmomatic setup is more complicated than the just pasting a specific name of adaptor. 
I made a mistake here assuming `ILLUMINACLIP:TruSeq3-PE.fa:2:30:10` is a name. but in realty in means:
```
1. Cut Illumina adaptor
2. The adaptor is the more common version of Truseq3, pair-ended
3.maximum mismatch count = 2
4.  The accuracy of the match between the two 'adapter ligated' reads must be for PE palindrome read alignment = 30.
5.  The accuracy of the match between any adapter etc. sequence must be against a read = 10
```
which makes them 5 parameters instead of 1.


The rule to extract such info from fastqc report, based on trimmonia, is as followed:
```
As a rule of thumb newer libraries will 
use TruSeq3, but this really depends on your service provider.
If you use FASTQC, the „Overrepresented Sequences‟ report can help indicate which adapter 
file is best suited for your data. 
“Illumina Single End” or “Illumina Paired End” sequences indicate single-end or paired-end 
TruSeq2 libraries, and the appropriate adapter files are “TruSeq2-SE.fa” and “TruSeq2-PE.fa” 
respectively. 
“TruSeq Universal Adapter” or “TruSeq Adapter, Index …” indicates TruSeq-3 libraries, and 
the appropriate adapter files are “TruSeq3-SE.fa” or “TruSeq3-PE.fa”, for single-end and 
paired-end data respectively. 
Adapter sequences for TruSeq2 multiplexed libraries, indicated by “Illumina Multiplexing 
…”, and the various RNA library preparations are not currently included.
```
One thing to that catch my attention is `the various RNA library preparations are not currently included`, so I am not sure as in what adapter option to use

hsun3163: There is the scenario where none of the aforementioned tags are reported. For example, in our provided dataset, following are the `Overrepresented Sequences` results, no hit means none of the known Truseq adaptors are there. Which may hints toward the  `the various RNA library preparations are not currently included` case.

```
>>Overrepresented sequences     fail
#Sequence       Count   Percentage      Possible Source
NNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN     158     1.58    No Hit
GCTCAGGCTGGAGTGCAGTGGCTATTCACAGGCGCGATCCCACTACTGAT      29      0.29    No Hit
CTCAGGCTGGAGTGCAGTGGCTATTCACAGGCGCGATCCCACTACTGATC      25      0.25    No Hit
GCCGGGTTGAATCCTCCGGGCGGACTGCGCGGACCCCACCCGTTTACCTC      25      0.25    No Hit
GCCCTCTTGAACTCTCTCTTCAAAGTTCTTTTCAACTTTCCCTTACGGTA      22      0.22    No Hit
CCAGGCTGGAGTGCAGTGGCTATTCACAGGCGCGATCCCACTACTGATCA      19      0.19    No Hit
CAGGCTGGAGTGCAGTGGCTATTCACAGGCGCGATCCCACTACTGATCAG      18      0.18    No Hit
GGCTGGAGTGCAGTGGCTATTCACAGGCGCGATCCCACTACTGATCAGCA      17      0.16999999999999998     No Hit
GTCGGGTCTGCGAGAGCGCCAGCTATCCTGAGGGAAACTTCGGAGGGAAC      16      0.16    No Hit
CCCCTCCTTAGGCAACCTGGTGGTCCCCCGCTCCCGGGAGGTCACCATAT      13      0.13    No Hit
CCCTCCTTAGGCAACCTGGTGGTCCCCCGCTCCCGGGAGGTCACCATATT      12      0.12    No Hit
CCTTAGGCAACCTGGTGGTCCCCCGCTCCCGGGAGGTCACCATATTGATG      12      0.12    No Hit
CCCTCTTGAACTCTCTCTTCAAAGTTCTTTTCAACTTTCCCTTACGGTAC      12      0.12    No Hit
GGTTCTTCGAATGTGTGGTAGGGTGGGGGGCATCCATATAGTCACTCCAG      11      0.11    No Hit
```

gaow: @hsun3163 thank you for the detailed investigation on this. Very enlightening. So indeed we should make all of them parameters, taken from http://www.usadellab.org/cms/?page=trimmomatic

```
Step options:

ILLUMINACLIP:<fastaWithAdaptersEtc>:<seed mismatches>:<palindrome clip threshold>:<simple clip threshold>
fastaWithAdaptersEtc: specifies the path to a fasta file containing all the adapters, PCR sequences etc. The naming of the various sequences within this file determines how they are used. See below.
seedMismatches: specifies the maximum mismatch count which will still allow a full match to be performed
palindromeClipThreshold: specifies how accurate the match between the two 'adapter ligated' reads must be for PE palindrome read alignment.
simpleClipThreshold: specifies how accurate the match between any adapter etc. sequence must be against a read.
SLIDINGWINDOW:<windowSize>:<requiredQuality>
windowSize: specifies the number of bases to average across
requiredQuality: specifies the average quality required.
LEADING:<quality>
quality: Specifies the minimum quality required to keep a base.
TRAILING:<quality>
quality: Specifies the minimum quality required to keep a base.
CROP:<length>
length: The number of bases to keep, from the start of the read.
HEADCROP:<length>
length: The number of bases to remove from the start of the read.
MINLEN:<length>
length: Specifies the minimum length of reads to be kept.
```

However only make `fastaWithAdaptersEtc` required parameter, and  tell people that this is a file name and it can be generated using fastqc (?). Other parameters should still be explicit parameters but we put in reasonable default values. The user interface will then require `--fastaWithAdaptersEtc` and nothing else.

hsun3163:
Alternativily, we can use the Trim Galore tool which more or less auto detect the adaptor and save us the trouble. We can keep trimmonia for everything else.

https://github.com/FelixKrueger/TrimGalore/blob/master/Docs/Trim_Galore_User_Guide.md

gaow:
> Alternativily, we can use the Trim Galore tool which more or less auto detect the adaptor and save us the trouble. We can keep trimmonia for everything else.

so you want to use this for the fastqc wrapper part? It's just a perl script ... https://github.com/FelixKrueger/TrimGalore/blob/master/trim_galore and it seems to run fastqc "after" trimming with `Cutadapt `. I dont think it is a good idea. 

However it might be a good idea to include fastqc into the container and run it prior to trimmonia. We can code it such that `fastaWithAdaptersEtc` be set automatically -- as the output from fastqc step.

hsun3163:
> > Alternativily, we can use the Trim Galore tool which more or less auto detect the adaptor and save us the trouble. We can keep trimmonia for everything else.
> 
> so you want to use this for the fastqc wrapper part? It's just a perl script ... https://github.com/FelixKrueger/TrimGalore/blob/master/trim_galore and it seems to run fastqc "after" trimming with `Cutadapt `. I dont think it is a good idea.
> 
> However it might be a good idea to include fastqc into the container and run it prior to trimmonia. We can code it such that `fastaWithAdaptersEtc` be set automatically -- as the output from fastqc step.

fastqc is already within the container and there is a step dedicated to it to generate the aforementioned html report.

pipeline/molecular_phenotypes/bulk_expression.ipynb
hsun3163: The problem is, as outlined in the example, sometimes we cant tell what the adapter is by fastqc though.

gaow: Sure. And if that is the case we should automatically skip the trim. @floutt I will pin you tomorrow on the other repo to ask about the data.

hsun3163:
The new ref_processing module is done:
1. gff3 to gtf conversion
2. collapsed of gtf file
3. Unify the chr format for fasta and gtf
4. STAR_index generation
5. RSEM_index generation 

TBD:
Test the remaining of bulk expression with the generated index.

PR #104, opened 2022-01-26
disable science notation in gtf

hsun3163:
no more issue with collapes_annotaion.py


gaow:
Great! @hsun3163 it's actually better if you implement that conversion using shell script eg `sed` instead of  reading it into R ... something I suggest you try in the future (use more bash and less R)

PR #119, opened 2022-02-01
sample merger & readcount_QC NA warning done

hsun3163:
The output of rnaseqc sample merger can run within readcount_QC up to a point where mahalanobis  was calculated to identify outlier.
```
Error in solve.default(cov, ...) :
  Lapack routine dgesv: system is exactly singular: U[1,1] = 0
Calls: mahalanobis -> solve -> solve.default
Execution halted
```
The reason for this error is that there the three samples are too similar to each other, so that the covariance matrix of genes are all zero.

However, the fact that the output of readcount_QC can pass the data integrity check indicates that the merging is successful


gaow:
Thanks @hsun3163 I believe that issue will be resolved with updated MWE (@floutt ).

PR #122, opened 2022-02-02
RSEM output aggregate

hsun3163:
for the aggregate_rsem_result.py, it will produce 1 such file for all the genes&samples :
```
gene_id transcript_id(s)        samp1   samp2
gene:ENSG00000000003    transcript:ENST00000373020,transcript:ENST00000494424,transcript:ENST00000496771,transcript:ENST00000612152,transcript:ENST00000614008  0       0
gene:ENSG00000000005    transcript:ENST00000373031,transcript:ENST00000485971   0       0
gene:ENSG00000000419    transcript:ENST00000371582,transcript:ENST00000371584,transcript:ENST00000371588,transcript:ENST00000413082,transcript:ENST00000466152,transcript:ENST00000494752       0       0
```
for each of the columns in the rsem output.  Which I think defeat our purpose in archiving the results but may actually be easier for the group to use considering its similarity with bed/gct format

### DIFF
--- a/pipeline/molecular_phenotypes/calling/RNA_calling.ipynb
+++ b/pipeline/molecular_phenotypes/calling/RNA_calling.ipynb
@@ -187,7 +187,7 @@
     "    --STAR-index reference_data/STAR_Index/ \\\n",
     "    --RSEM-index reference_data/RSEM_Index/ \\\n",
     "    --container container/rna_quantification.sif \\\n",
-    "    --mem 40G --is_stranded True"
+    "    --mem 40G --is_paired_end &"
    ]
   },
   {
@@ -351,6 +351,8 @@
     "parameter: data_dir = path(f\"{samples:d}\")\n",
     "# Is the RNA-seq data stranded\n",
     "parameter: is_stranded = False\n",
+    "# Is the RNA-seq data pair-end\n",
+    "parameter: is_paired_end = False\n",
     "# For cluster jobs, number commands to run per job\n",
     "parameter: job_size = 1\n",
     "# Wall clock time expected\n",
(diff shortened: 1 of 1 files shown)


pipeline/molecular_phenotypes/calling/RNA_calling.ipynb
```diff
@@ -187,9 +187,19 @@
     "    --STAR-index reference_data/STAR_Index/ \\\n",
     "    --RSEM-index reference_data/RSEM_Index/ \\\n",
     "    --container container/rna_quantification.sif \\\n",
-    "    --mem 40G"
+    "    --mem 40G --is_stranded True"
```
gaow, line 190: Is our data stranded? When preparing for the reference i didn't use stranded option ...

Also please use `--is-stranded` without `True` to flag this. it's a boolean variable in SoS

pipeline/molecular_phenotypes/calling/RNA_calling.ipynb
```diff
@@ -714,6 +725,64 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8a2b8cf8-b9a1-4440-a86f-3fbbb610e61b",
+   "metadata": {
+    "kernel": "SoS"
+   },
+   "outputs": [],
+   "source": [
+    "[rnaseqc_call_5]\n",
```
gaow, line 749: Do we need to use the aggregate rnaseqc result script from Broad?

pipeline/molecular_phenotypes/calling/RNA_calling.ipynb
hsun3163: Fixed, didnt realized some changes was not saved.

pipeline/molecular_phenotypes/calling/RNA_calling.ipynb
hsun3163: No we dont need to, but I thought it was preferred to use whatever they have for the pipeline?

gaow: Yes, it's preferred to use their code

PR #136, opened 2022-02-09
Gene_cood annot fix, normaization tested, bioinfo docker fix

hsun3163:
1. Gene_coord is fixed and tested with the mwe made from log2cpm
1.1. A  `${\"None\" if not sample_participant_lookup.is_file() else \"pd.read_csv(sample_participant_lookup, sep='\\t', index_col=0, dtype={0:str,1:str}, squeeze=True)\"}\n"` statement was replaced since it will cause parsing error, possible due to too many quotation marks

2. -o was added to dockerfile of bioinfo to overwrite an error arise from pre-existed plink in /tmp/ folder (reason unknown)
3. Change the biopython in bioinfo to LDtools in accordance to the need of sumstat_merger module


code/data_preprocessing/phenotype/gene_annotation.ipynb
```diff
@@ -94,10 +94,10 @@
    "source": [
     "sos run gene_annotation.ipynb annotate_coord \\\n",
     "    --cwd output \\\n",
-    "    --phenoFile data/... \\\n",
-    "    --annotation-gtf reference_data/Homo_sapiens.GRCh38.103.chr.reformatted.gene.ERCC.gtf  \\\n",
+    "    --phenoFile data/MWE.pheno_log2cpm.tsv \\\n",
+    "    --annotation-gtf data/gene.gtf  \\\n",
```
gaow, line 98: can we use our `Homo_sapiens.GRCh38.103.chr.reformatted.gene.ERCC.gtf` for MWE purpose and not upload another GTF?

code/data_preprocessing/phenotype/gene_annotation.ipynb
```diff
@@ -94,10 +94,10 @@
...
     "    --sample-participant-lookup data/sampleSheetAfterQC.txt \\\n",
-    "    --container container/rna_quantification.sif \\"
+    "    --container ./rna_quantification.sif --phenotype_id_type gene_name"
```
gaow, line 100: you might have noticed my convention to use eg `--phenotype-id-type` and not `--phenotype_id_type` in command calls. Please stick to my convention to be consistent.

gaow:
>  statement was replaced since it will cause parsing error, possible due to too many quotation marks

I suspect it's nested `{}` and solution might be to change `{0:str, 1:str}` to `dict([(0,str),(1,str)])`. Do you want to try that? It's best to know *exactly* what went wrong and not *possibly* what went wrong.

hsun3163:
> > statement was replaced since it will cause parsing error, possible due to too many quotation marks
> 
> I suspect it's nested `{}` and solution might be to change `{0:str, 1:str}` to `dict([(0,str),(1,str)])`. Do you want to try that? It's best to know _exactly_ what went wrong and not _possibly_ what went wrong.

I just tried and this is not the reason...in fact I remember using nested {} before via f string and those are fine.

code/data_preprocessing/phenotype/gene_annotation.ipynb
hsun3163: Just tried and it works, reverted.

gaow:
@hsun3163 the parsing error is because f-string in Python does not accept `\` ... I'll merge your version for now and make some changes on that, and push back

PR #141, opened 2022-02-10
Apex mwe + sample_matcher

hsun3163:
The snbu data analysis got to a point where the proposed sample matching mechanism needs to be done.
1. It is a misc module and intended to be used repeatly
1.1 1 times on plink the output of vcf qc
1.2 2 times on the gct.gz output of bulk_expression_qc
1.3 1 time on the output of merged_plink after GWAS_QC ?  

2. Only thing it do is remove samples from input data, gct like for phenotype and plink for genotype
2.1 sample renaming will be done to by downstream module, normalization/annotation. 

3. No need to run it on cov, Both factor analysis shall only analysis the sample in phenotype file

**4. Need to clarify: Is the sample_lookup table static or will be change back and forth, i.e after pca ?**

### DIFF
--- a/code/data_preprocessing/covariate/BiCV_factor.ipynb
+++ b/code/data_preprocessing/covariate/BiCV_factor.ipynb
@@ -69,16 +69,6 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "aa73d907-b6c6-4604-b792-aacdffc3455f",
-   "metadata": {
-    "kernel": "SoS"
-   },
-   "outputs": [],
-   "source": []
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
    "id": "impressed-photographer",
    "metadata": {
     "kernel": "SoS"
--- a/code/data_preprocessing/covariate/BiCV_factor.ipynb
+++ b/code/data_preprocessing/covariate/BiCV_factor.ipynb
@@ -20,11 +20,22 @@
    "source": [
     "## Overview\n",
     "\n",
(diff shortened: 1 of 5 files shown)


gaow:
@hsun3163 thanks -- at this point having the code in place to match sample is good enough. I have other thoughts about not making it a `misc` module but we can discuss. Maybe i'm not thinking it through

> 1.1 1 times on plink the output of vcf qc

We have a `--keep-samples` option in KING pipeline I think we can use it directly to extract the subset of individuals as well as separating between related and unrelated.

>  1.2 2 times on the gct.gz output of bulk_expression_qc

yes, on the read count and on the TPM. But I"m more thinking of it as a second QC step as it essentially boils down to removing samples

> 1.3 1 time on the output of merged_plink after GWAS_QC ?

this we can discuss -- I dont think it is necessary to create a new genotype file per molecular phenotye.

> Both factor analysis shall only analysis the sample in phenotype file

Need to make sure it is the case for PEER. 

> 4. Need to clarify: Is the sample_lookup table static or will be change back and forth, i.e after pca ?

Indeed outliers may come from PCA on genotype as well as from phenotype QC eg those based on clustering of TPM , and they should be constantly removed from genotype and phenotype ensure everything matches. It is going to be very tedious. I think the reason we need to match is to make sure that PCA, normalization and PEER results are "roughly" correct, and i don't think these result is going to change that much with a few outliers involved. I therefore believe it is reasonable to allow for a few outliers here and there not completely removed prior to association analysis, and association analysis modules should be able to drop samples with missing genotype or phenotype data.

Issue #145, opened 2022-02-11
Engineering problem for implementing parallelization for VCF_QC


hsun3163:
Goal: conduct vcf_qc per chrome

Features:
1.  VCF_QC should be able to handle 1 file vcf input and a list file of vcf input
2.  VCF_QC should be able to use vcf with chr name and without, vcf without chr name shall be preprocessed first

## Proposed mechanism 1:
```
[dbsnp_1, qc_1, rename_chr]
skip_if(rename_chr = False)

[qc_2, (variant preprocessing)]
input: f'{output_from("X") if rename_chr else genoFile}'
```
This design didnt work, give an error message as follow:
```
Invalid directive: SyntaxError('invalid syntax', ('<string>', 1, 10, 'def func(f\'{test_file if Skip = False else output_from("A")}\'\n'))
```

## Proposed mechanism 1.1:

similar to 1 put the if statement outside of input:

```
[A]
skip_if(skip)
input: test_file, group_by = 1
output: f'{_input}.mod'
bash: expand= "${ }"
    echo "modded" >> ${_output}

[B]
input: output_from("A") , group_with = "test_file"
input_file = _test_file if skip else _input 
bash: expand= "${ }"
    echo ${input_file}
```

Expected behavior:
When skip = True, B will works on each of the _test_file
When skip = False, B will works on each of the output from A

Actual behavior:
When skip = False. it works
When skip = True, following error occurs:
```
INFO: Running A:
INFO: Running B:
ERROR: [B]: [B]: Failed to process input statement output_from("A") , group_with = "test_file"
: Length of provided properties (2) does not match number of groups (1)
```
Which should be due to output_from A being empty

## Proposed mechanism 2:

Ditch the fancy mechanism, but do
```
[qc_add_chr_1, rename_chr]
...
[qc_add_chr_2, variant preprocessing_add_chr ]

[qc_1, variant preprocessing]

```

where [qc_1, variant preprocessing] and [qc_add_chr_2, variant preprocessing_add_chr ] is enssentially the same block of codes with only the input differs. Proposed mechanism 2 should definitly work for it is nothing new.

## Propose mechanism 3:
Use a wrapper to run vcf_qc, which also will definitly work but not fit into our overall design philosophy.

hsun3163:
The ipynb I used to tested mechanism 1.1 is attached. Due to restriction of github it was gzipped
[...]

PR #146, opened 2022-02-12
PEER + pca done

hsun3163:
Cluster getting super slow so send in the Daily PR before cluster stuck
1. PEER done
2. Changes in vcf_qc not implemented yet, still trying the "provides" scripts suggested in #145 
3. As mentioned in slack, need a way to check the output vcf.gz file in qc_1. Its size  ~1.07 times that of the input file which is consistant with previouse test example. But need to make sure
4. Also needs to add --threads option to all the qc step to speed up the analysis. I am not familiar with this operation. How is --threads 10 looks like?

### DIFF
--- a/code/data_preprocessing/genotype/VCF_QC.ipynb
+++ b/code/data_preprocessing/genotype/VCF_QC.ipynb
@@ -75,8 +75,8 @@
    },
    "source": [
     "## Input\n",
-    "\n",
-    "1. The target `VCF` file\n",
+    "a list \n",
+    "1.1. The target `VCF` file\n",
     "    - If its chromosome name does not have the `chr` prefix and you need it to match with reference `fasta` file, please run `rename_chrs` workflow to add `chr`.\n",
     "2. dbSNP database in `VCF` format\n",
     "3. A reference sequence `fasta` file"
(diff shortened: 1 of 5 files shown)


hsun3163:
1. VCF_QC parallel done, dryrun can run
2. vcf_qc ID fix, now RSID in info and ID is chr:pos:alt:ref, tested with mwe
3. genotype_formating get_genotype_function bugfix, same function as that of vcf_qc

code/data_preprocessing/genotype/VCF_QC.ipynb
```diff
@@ -456,10 +540,11 @@
     "task: trunk_workers = 1, trunk_size=job_size, walltime = walltime, mem = mem, cores = numThreads, tags = f'{step_name}_{_output:bn}'\n",
     "bash: container = container, expand= \"${ }\", stderr = f'{_output:n}.stderr', stdout = f'{_output:n}.stdout'\n",
     "        ${'bcftools norm -m-any' if not bi_allelic else 'bcftools view -m2 -M2'} ${'-v snps' if snp_only else \"\"} ${_input} |\\\n",
-    "        bcftools norm --check-ref w -f ${reference_genome}  -Oz|\\\n",
+    "        bcftools norm -d exact -N --check-ref w -f ${reference_genome}  -Oz --threads ${numThreads} |\\\n",
```
gaow, line 543: I just noticed there are two `bcftools norm` commands here. Will the first `norm` perform the indel normalization we dont like? Or it's fine since `-f` is not provided? I'm not sure why we need two `norm` anyways. @hsun3163 please document what each command does if you review and decide this is the right thing to do.

PR #158, opened 2022-02-17
fix on get genotype and grm

hsun3163:
Unify the get_genotype_function from vcf_qc, grm and genotype_reformatting.

minimum changes was applied to the function.


code/data_preprocessing/genotype/VCF_QC.ipynb
```diff
@@ -377,10 +377,10 @@
    "source": [
     "[global]\n",
     "# input can either be 1 vcf genoFile, or a list of vcf genoFile.\n",
-    "parameter: genoFile = path\n",
+    "parameter: genoFile = paths\n",
     "\n",
     "# Workdir\n",
-    "parameter: cwd = path\n",
+    "parameter: cwd = paths\n",
```
gaow, line 383: this should be `path` not `paths`

hsun3163: if we expect it to use the same function as genotype_reformatting, then this need to be paths...

hsun3163: Oh nvm i did see it is cwd, indeed this should be path, changed

PR #197, opened 2022-03-21
sumstat_merger_split

hsun3163:
now sumstat merger is two step: unify sumstat and sumstat to vcf, as discussed.


hsun3163:
1. yml generator now generate yml properly and based on pre-defined templates
2. sumstat merger unifier take target as a 5 columns index file instead of a sumstat file, as just discussed

PR #198, opened 2022-03-24
Added git to the dockerfile

floutt:
One of the packages `gffread` requires Git for its installation. I ran into some issues when running this and fixed it by added the `git` package to the apt command and rerunning this.


gaow:
Thank you @floutt I have updated the corresponding singularity file `container/singularity/rna_quantification.def`

Issue #199, opened 2022-03-24
NA in covariates when removing covariates via PEER's remove covariates functions.


hsun3163:
In the code provided by peers.
The covariates were removed based on a estimated QR:
```
# In this code, we'll repeatedly regress different Y against the same X
  # (covariates) in order to calculate residuals.  Rather that repeatedly
  # call lm to do this, we'll avoid recalculating the QR decomposition for the
  # covariates matrix each time by reusing it after calculating it once
```
When there is NA values in the covariates of one sample, QR for that particular sample will not exist. Preventing the covariates removal to be done.

This can be "fixed" by removing the samples with missing covariates. However, it is uncleared whether doing so is a good decision.

_____________________________________________________________________________
PS. In our orginal workflow, samples will not have missing covariates, as those will be filled by APEX/PEER, and no needs to remove .

PSS. The main feature from the PEER's remove covariates is that they calculate the residual from each gene expression, and use the QR while doing so.

hsun3163:
For missing covariates, we have introduced 3 modes to handle them. See details in covariate_formatting notebook

PR #202, opened 2022-03-30
PEER testing done, sumstat merger ongoing

hsun3163:
1. PEER testing done
2. Remove covariate testing done, for #199 now covariates with NA were removed.
3. testing of sumstat merger require building of 1 whole genome "TARGET" or 23 "TARGET"s. Ongoing


gaow:
> now covariates with NA were removed.

what does this mean -- did you delete this covariate all together, or you use mean imputation for the covariate value before regressing it out?

Issue #218, opened 2022-04-15
All na result in APEX


hsun3163:
The output of APEX produce all na result.

Which may be due to following LMM output:
```
8       232136  232137  ZNF596  nan     nan     3819.66 nan     0.991691
8       406427  406428  FBXO25  nan     nan     3819.66 nan     0.986391
8       545780  545781  TDRP    nan     nan     3819.66 nan     0.979933
8       737595  737596  DLGAP2  nan     nan     3819.66 nan     0.985495
8       738105  738106  ERICH1  nan     nan     3819.66 nan     0.985181
8       1622416 1622417 DLGAP2-AS1      nan     nan     3819.66 nan     0.987079
8       1755777 1755778 CLN8    nan     nan     3819.66 nan     0.992123
8       1764507 1764508 AC100810.1      nan     nan     3819.66 nan     0.988725
8       1823925 1823926 ARHGEF10        nan     nan     3819.66 nan     0.988338
8       1973676 1973677 KBTBD11 nan     nan     3819.66 nan     0.992292
```
It is yet to determined 1. what each of the columns are and 2. why nan in here.

Another clue is this in the  production log it said decomposing 0 block out of 1:
```
sing 1 threads.
324887 total variants on selected chromosomes.

Found 1194 samples in bcf file ...
Found 415 samples in covariate file ...
Found 424 samples in expression bed file ...
Found 415 samples in common across all three files.

Processed data for 86 covariates across 415 samples.
Processed expression for 482 genes across 415 samples.
Found 1 related blocks with 415 total individuals (largest block = 415) ...
Reordering related GRM blocks ... Done.
Decomposing block 0 (415) out of 1 ... Donek
.Creating eigenvector matrix ... Done.
```

hsun3163:
The problem rooted in the GRM file as shown in the following log:
```
Reading the PLINK FAM files ....
1194 individuals have been included from the PLINK FAM files.
Reading the PLINK BIM files ...
Warning: Duplicated SNP ID  has been changed to chr1:0:00.
Warning: Duplicated SNP ID  has been changed to chr1:0:00.
Warning: Duplicated SNP ID  has been changed to chr1:0:00.
Warning: Duplicated SNP ID  has been changed to chr1:0:00.
Warning: Duplicated SNP ID  has been changed to chr1:0:00.
Warning: Duplicated SNP ID  has been changed to chr1:0:00.
Warning: Duplicated SNP ID  has been changed to chr1:0:00.
[...]

PR #219, opened 2022-04-16
add UniSuSiE to command generator

hsun3163:
Several bug fix and addition of UniSuSiE to command generator (WIP)

May change the UniSuSIE module instead if current implementation proven to be not good, namely partitioning genotype based on  the region list of multiple tissue, may be concatenation of region list is needed.


hsun3163:
Implement fix for #218

PR #238, opened 2022-05-04
improvement on the release.sos

hsun3163:
1. Modify the release.sos to document the output and error message of singularity build.
2. Add fastp to rna_quant
3. update STAR to 2.7.10a
4. fix a bug in the release.sos so that the singularity build can be ran on any location.


release.sos
```diff
@@ -12,9 +12,9 @@ bash: expand = True
 [singularity]
 parameter: config = path
 parameter: out_dir = path('container/singularity')
-bash: expand = True
+bash: expand = True, stderr = f'{config}.stderr', stdout = f'{config}.stdout'
   rm -f {out_dir}/{config:n}.sif
-  singularity build --fakeroot {out_dir}/{config:n}.sif container/singularity/{config}
+  singularity build --fakeroot {out_dir}/{config:n}.sif {config}
```
gaow, line 17: in this case we should use `{out_dir}/{config:bn}.sif` otherwise you'll end up with the path of config file inside `out_dir`

gaow: Also please add to gitignore to skip these files:

```
**/*.sif
**/*.stdout
**/*.stderr
```

hsun3163:
1. Step 6 of the QC code implemented:
1.1 It is simply running the tool multiQC to scan through the working directory to generate a QC report

2. Step 7 pending
2.1 Step 7 is to merge qc metric from picard and RSEM to generate a QC table. However, the required input: `*_rsem.ctn` and `duplicate_metrics` folder cant be found.

PR #244, opened 2022-05-11
Replaced --stranded flag with --is-stranded

floutt:
Minor error. Replaced run_RSEM flag `--stranded` with `--is_stranded`


gaow:
@floutt we are doing some edits here on our end to for the strand issue -- I am wondering if you actually know in your data is it FR or RF stranded? [See this project](https://github.com/betsig/how_are_we_stranded_here) for more details , and do you need to run something like this for your project, or you know the exact library type to tell the strandedness already?

PR #245, opened 2022-05-11
Removed redundant hard set of paired end parameter

floutt:
I noticed I was getting errors when running this pipeline and was getting errors involving paired end reads from RSEM. Looking at the code here it appears that the paired end parameter was statically set as false despite being dynamically defined prior to this. To get around this I deleted the statically defined parameter `paired_end` (since it was redundant as there was already a paired end parameter) and replaced it with the `is_paired_end` parameter.


floutt:
Okay, sounds good!

PR #250, opened 2022-05-13
APEX MWE and rnaseqc tested

hsun3163:
The reason `--fit-null` are forgone are listed in the notebook, basically the three were replaced by other commands. There was a interface change be fore we try out neuro-apex and the xqtl project.


gaow:
@hsun3163 good documentation on APEX. Thanks!

PR #251, opened 2022-05-13
TensorQTL MWE

hsun3163:
1. Remove for_each
2. Add filter by gene
3. fix trans


hsun3163:
The conflict is likely due to the insertion of new image, which I think the simplest way may be remove it in this repo and add the new one back.

gaow:
@hsun3163 I did what you suggested above. Please refork.

PR #260, opened 2022-05-20
strand detection pending

hsun3163:
1. confirmed by infer_experiment.py that our MWE is indeed RF, as shown below:
```
 infer_experiment.py  -r stranded_test_samp1_r1/Homo_sapiens.GRCh38.103.chr.reformated.bed  -i ./output/sample_1.Aligned.sortedByCoord.out.bam
[W::hts_idx_load3] The index file is older than the data file: ./output/sample_1.Aligned.sortedByCoord.out.bam.bai
Reading reference gene model stranded_test_samp1_r1/Homo_sapiens.GRCh38.103.chr.reformated.bed ... Done
Loading SAM/BAM file ...  Finished
Total 14835 usable reads were sampled

This is PairEnd Data
Fraction of reads failed to determine: 0.1056
Fraction of reads explained by "1++,1--,2+-,2-+": 0.0254
Fraction of reads explained by "1+-,1-+,2++,2--": 0.8690
```
1.1 how_are_we_stranded not working, but the bed file it generated are used in producing 1
1.2 The above result is in fact contaminated by "how_are_we_stranded" standard.

2. Implemented the mechanism of detecting `strand` and replacing `strand` with user input.
2.1 The percentage that corresponds to "rf" is, in fact, "Counts for the 2nd read strand aligned with RNA is over 90% of aligned count", which is confusing as here it said **2nd read strand** but rf is `Data is likely RF/fr-firststrand` by "how_are_we_stranded " standard.
2.2 Our MWE produces an RF using 2, consistent with 1.

3. At this point, each substep for each sample produces a `strand` variable. Unknown how to unify them and share with another step yet, perhaps a step with `group_by = "ALL"` is needed.


hsun3163:
Please don't merge it to the main branch yet. It is likely to break.

hsun3163:
Point 3 is fixed. Dryrun test passed. Actual run test ongoing. Since there are no changes to the actual analysis code, the actual run should also pass.

code/molecular_phenotypes/calling/RNA_calling.ipynb
```diff
@@ -433,6 +429,7 @@
     "cwd = path(f'{cwd:a}')\n",
     "# Reference gene model\n",
     "parameter: gtf = path\n",
+    "parameter: RSEM_index = path\n",
```
gaow, line 432: this should not be a global parameter. It should be in RSEM step

gaow: Is there a reason you must do this instead?

hsun3163: It was a leftover from when I debugged, fixed it.

PR #269, opened 2022-05-24
bug fix and mwe of leafcutter, update on phenotype/covariate/genotype tutorial page

hsun3163:
1. The existed interface/module of leafcutter is mostly functional with the MWE
1.1 a fixed bug that was fixed in the leafcutter container cause the last step not working (new sif building)
1.2 The MWE can not produce correct result. The bam used to produce the correct result is 25G.
1.3 Leafcutter now take the output of star_align step of rna_calling module, this could be changed if we want to refactor them.

2. Update the command section with the codes generated by the command generator.


code/molecular_phenotypes/calling/splicing_calling.ipynb
```diff
@@ -306,7 +309,7 @@
     "task: trunk_workers = 1, trunk_size = job_size, walltime = walltime, mem = mem, cores = numThreads\n",
     "bash: expand= \"${ }\", stderr = f'{_output[0]:n}.stderr', stdout = f'{_output[0]:n}.stdout', container=container\n",
     "    cat ${_input} > ${_output:nn}.junc\n",
-    "    leafcutter_cluster_regtools.py -j ${_output:nn}.junc -m ${min_split_reads} -o ${_output:nn} -l ${max_intron_len}"
+    "    python leafcutter_cluster_regtools.py -j ${_output:nn}.junc -m ${min_split_reads} -o ${_output:nn} -l ${max_intron_len}"
```
gaow, line 312: in the docker image, we should copy this script to `/usr/local/bin` and `chmod +x` then use it without having to type `python`

container/singularity/leafcutter.def
```diff
@@ -38,6 +42,8 @@ tar -xf samtools-1.11.tar.bz2 && rm samtools-1.11.tar.bz2 && cd samtools-1.11 &&
 export PATH=/opt/samtools-1.11:$PATH
 export PATH=/opt/htslib-1.11:$PATH
 export PATH=/opt/regtools/build:$PATH
+export PATH=/opt/leafcutter/clustering:$PATH
+alias leafcutter_cluster_regtools.py="python leafcutter_cluster_regtools.py"
```
gaow, line 46: let's not do this -- instead use the approach as I commented above

code/molecular_phenotypes/calling/splicing_calling.ipynb
hsun3163: As it turns out, I actually don't need to chmod + x and add the python explicitly. will remove the alias statement and the python in the leafcutter.ipynb

Issue #265, opened 2022-05-23
Output name for merge_factor step in covariate_formatting.


hsun3163:
The idea of having `{cwd:a}/{factorFile:bnn}.{covFile:bn}.cov.gz` as output make sense. However, when stacking with the previous naming convention, the output file after merge_pca, compute_resid, PEER, and merge_factor will become:

`MWE.log2cpm.MWE.covariate.cov.MWE.MWE.related.filtered.extracted.pca.projected.resid.bed.PEER.MWE.covariate.cov.MWE.MWE.related.filtered.extracted.pca.projected.cov.gz`

which is `{phenoFile}.{covFile}.{pcaFile}.resid.PEER.{covFile}.{pcaFile}.cov.gz`

I understand that the current implementation is the most comprehensive in tracking the footprint of the file; just wondering if it is indeed necessary.

gaow:
We have allowed for `name` to be specified so people can overwrite this default. Still this default can certainly be improved ... 

1. Factor analysis output default name `{phenoFile}.{covFile}.{pcaFile}.resid.PEER` is reasonable, if we can make that happen.
2. We need to get rid of `bed` in the filename.

Can we make sure these changes happen to the default ; then in the documentation you can also show people how to use --name to overwrite it?

hsun3163:
The default in the command generator will overwrite the merge_pca name and the merge_covariate name. As the pca name inherited the genotype file name and lumping together will cause an OSError 36 in tensorQTL.

PR #270, opened 2022-05-25
variouse fixes on the command generator's covariates and genotypes

hsun3163:
1. Implemented name overwrite for all covariate step to not include the pcas file name as mentioned in #265 
2. Various bug fix for the genotype session
3. Add a if statement so that the recipe does not needs to be specified if the genotype of command generator was ran
4. Fix a bug in the leafcutter
5. Add the path `python {path}` as a temporary solution until the PR for the leafcutter is accepted.


hsun3163:
As it turn out, there is an tricky problem introduced by the leafcutter.py. Further diagnosis needed.

PR #274, opened 2022-05-27
Edited splicing_calling.ipynb and current one is usable

Rhopala:


gaow:
@Rhopala sorry your edits look good but there is a merge conflict. While it is easy for me to fix it, I would rather have you try and understand what it is, and know how to handle it in the future. Thank you!

Rhopala:
@gaow I think it's solved now.

Issue #277, opened 2022-06-01
removal of "*" symbol in sumstat by default.


hsun3163:
The analysis of the newly completed 8tissue analysis vcf.gz was blocked due to the following errors:
```
Error: scanVcf: _DNAencode(): invalid DNAString input character: '*' (byte value 42)
  path: /mnt/vast/hpc/csg/snuc_pseudo_bulk/eight_tissue_analysis/output/data_intergration/TensorQTL/ALL_Ast_End_Exc_Inh_Mic_OPC_Oli.log2cpm.bed.processed_phenotype.per_chrom_ALL_Ast_End_Exc_Inh_Mic_OPC_Oli.log2cpm.ALL_Ast_End_Exc_Inh_Mic_OPC_Oli.covariate.pca.resid.PEER.cov.chr9.norminal.cis_long_table.merged.vcf.gz
Execution halted
```

This is due to the present of * symbol in the REF column of the VCF. 

In the previous analysis. This is avoided by using a prepared TARGET file that has all the * in the ALT columns.

However, when shipped out the workflow, I think it is crucial to check if the REF columns has "*" as part of the qc in the sumstat merger, independent of the TARGET file so that this problem will not go unnoticed and impact subsequent analysis.

PR #278, opened 2022-06-01
landing page up to cisQTL rewrite, add * detection in ss merger

hsun3163:
1. implemented what was proposed in #277 
2. Finished landing page rework up to xqtl.
2.1 It is still uncleared what is `**FIXME: still need to add in a section for VCF formatted file processing after PCA**` in genotype_processing landing page. As I understand, after PCA, the outlier can be removed from the QC plinks genotype file and follow-up analysis is done based on this QCed plink: partition_by_chrom for TensorQTL, and plink_to_vcf for APEX. No VCF post-processing seems to be needed. 

2.2 Speaking of, I was thinking maybe we should remove outliers from the covariate-factor-PCA file because this is the smallest file and it will achieve the same degree of sample removal as removing from the genotype.


gaow:
> As I understand, after PCA, the outlier can be removed from the QC plinks genotype file and follow-up analysis is done based on this QCed plink

So we have to require that genotype data be converted to PLINK? I thought we discussed earlier that we'll sake to support VCF file in TensorQTL so we dont have to make these unnecessary conversions for WGS data. That's why I have that note there. 

> because this is the smallest file and it will achieve the same degree of sample removal as removing from the genotype.

My understanding is that we rely on sample name list to subset genotype/phenotype/covariates -- outlier removal can be coded into any of these post QC steps that involves list of samples. But we should remove outliers before factors are computed because we don't want to include outliers in those calculations.

code/association_scan/assoc_result_processing.ipynb
```diff
@@ -29,9 +29,9 @@
    },
    "outputs": [],
    "source": [
-    "sos run /home/hs3163/GIT/xqtl-pipeline/code/association_scan/assoc_result_processing.ipynb genome \\\n",
-    "    --vcf  `ls Target_test/*merged.vcf.gz` \\\n",
-    "    --padjust_method \"bonferroni\" -J 500 -c csg.yml -q csg -s build   &"
+    "sos run pipeline/assoc_result_processing.ipynb genome \\\n",
+    "    --vcf  `ls output/data_intergration/TensorQTL/*merged.vcf.gz` \\\n",
+    "    --padjust_method \"bonferroni\" -J 500 -c csg.yml -q csg &"
```
gaow, line 34: Please use `-` convention for `--padjust-method`. And remove `-c -q &`

PR #298, opened 2022-06-15
ss standardizer overhaul, sqtl adaptation of tensorQTL from gtex

hsun3163:
TBD:
1. check/fix dup for pseudo vcf
2. test the output ss with MWE
3. a mechanism to handle TARGET file, maybe remove from the yml but take our output by default?


hsun3163:
The changes in each unit were tested. But the process from tensorQTL to standardized output still needs testing.

Will move on to the mock run of TensorQTL with the intron cluster first. Then come back for the overall testing.

hsun3163:
Added the sqtl pipeline from gtex to suit our needs. It only works with leafcutter for the time being, need to develop the psi part after psi was done.

hsun3163:
Tested with xuanhe's output, the procedure works mechanically. But the cis.map_permutation and cis.map_cis function both produce NaN result with the given data.

PR #311, opened 2022-07-06
methylation calling module done. Fix issues with RNA_calling

hsun3163:
Add prototype for methylation module.

WIP;
Needs testing and documentation for the rationale.


gaow:
@hsun3163 I'll review and merge when you are done with it. There are files under `pipeline/` folder they should be cleaned up and not committed here -- you can fix  that in the next commit.

hsun3163:
Unlike what was described in the gdrive dockerfile and the email from Bandita. The package "IlluminaHUmanMehylationEPICanno.ilm10b4.hg19" can not be omitted or avoided by using the attached R package. For the presence of "IlluminaHUmanMehylationEPICanno.ilm10b4.hg19" package is needed by a function embedded deeply of `preprocessQuantile` which was needed for normalization.

Also, installing this package on our cluster failed, for it requires R 4.2.

Therefore, installing it to the docker is required, still waiting for that to be finished.

gaow:
Thanks! So we have to work with hg19 ? Is there a way to  call that bsed on our hg38 reference?

gaow:
Merged to reflect various changes based on feedback from Monday WG meeting.

PR #336, opened 2022-07-29
CAFHE, complete case analysis for each gene independently

hsun3163:
For CAFHE, any na in any of the input DF: including LD, beta, se, n_samples will cause an error:
```
mtrand.pyx in numpy.random.mtrand.RandomState.choice()

ValueError: probabilities contain NaN
```

Therefore, complete case analysis implemented as discussed in slack


hsun3163:
Now it will first remove SNPs that are having NA result in all the study first, then remove studies that have NAs

PR #380, opened 2022-09-07
Add some utility in phenotype processing.

hsun3163:
1. Since the filter (and imputation) of missingness can potentially be useful for other bed.gz file, it was made a separate snippet in phenotype ipynb
2. The bam subsetting and bam to fastq are sth we need to parallelize for making the MWE. So should be added to sos as well.


gaow:
I'm not sure where we should put the notebook dedicated for MWE but let's leave it here for now. I am thinking of another repo for the protocol paper where these things should belong.

PR #395, opened 2022-09-15
Add de-identifying option to xqtl_protocol_demo

hsun3163:
After taking a closer look at the codes, as it turned out, the code for de-identifying will not undermine the process of de-identifying as long as we keep the mapper to ourselves, and thus should not be stored elsewhere.


gaow:
Thanks @hsun3163 we might want to move it elsewhere to keep the message focused on the xQTL demo. I will make changes as I see fit down the road.

PR #442, opened 2022-10-07
command gerneator done

hsun3163:
I also get rid of the `psi_hg38_annotation` step from the tutorial (Not the actual code but the command) for it was not entirly clear how to download it from  https://www.genenames.org/download/custom/ and what is the purpose for this output

### DIFF
--- a/code/commands_generator/eQTL_analysis_commands.ipynb
+++ b/code/commands_generator/eQTL_analysis_commands.ipynb
@@ -106,9 +106,8 @@
     "kernel": "SoS"
    },
    "source": [
-    "| Theme      | molecular_pheno | genotype_file |covariate_file| phenotype_group (sQTL) |\n",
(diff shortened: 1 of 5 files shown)


code/commands_generator/eQTL_analysis_commands.ipynb
```diff
@@ -106,9 +106,8 @@
     "kernel": "SoS"
```
gaow, line 106: <pre class="ql-syntax" spellcheck="false">Line #6.        --Association_option "TensorQTL" --cwd output2 --run  --yml "" --N 3   
</pre><p>what does it mean by <code>--yml  ""</code> ? if  the default is empty string then it should not need to be specified ... </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cumc/xqtl-pipeline/pull/442/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

PR #451, opened 2022-10-17
change naming mechanism to further avoid name too long error

hsun3163:
also update genotype processing with accurate Mem and time( pull from cluster records)


code/data_preprocessing/genotype/genotype_formatting.ipynb
```diff
@@ -388,7 +388,7 @@
     "[plink_by_gene_2]\n",
```
gaow, line 388: <p>this is not intended right?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cumc/xqtl-pipeline/pull/451/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

hsun3163: This is intended due to #454, I have tried to reproduce the error consistently but fail, so I decided to patch it this way as they don't need to be run with the task statement in the first place.

hsun3163:
1. Both softimpute and residualization can be done on the methylation input for the whole genome. Therefore softimpute were implemented as whole genome process and residualization are unchanged
2. add softImpute package to the methylation container
3. various other minor changes to simplify the filenames

PR #457, opened 2022-10-20
minor updates

hsun3163:
Will update the command for methylation later tonight


hsun3163:
Methylation instruction added.

PR #461, opened 2022-10-27
update with command generator

hsun3163:
1. Overhaul the mini protocol with the command from command generator
2. Remove the wrong Q value implemented previously(Chrom-wide instead of genome wide)
3. The new output from susie.


hsun3163:
Merge so that the fix on #465 can be used

PR #468, opened 2022-11-16
Adding if statement to deal with fewer than two columns in dat$Z

floutt:
When trying to run the code for susie-rss, I noticed that I was getting a strange error which said that I was trying to access an invalid index of `susie_tb_ls`. I then noticed that this was due to the assumption that there are multiple columns for the `dat$Z` matrix. In order to deal with this I added a conditional, so that this portion of the code only runs if the `susie_list` list has more than one element. This allowed for me to bypass that error


gaow:
Thanks @floutt this makes sense to me although I suspect there might be a more elegant way to do it. @hsun3163 could you take a look and if you think it is proper merge it? If you have other improvements you can make that after you merge.

PR #473, opened 2022-11-29
Some additions to pipelines

dianacornejo:


gaow (commented):
Thanks @dianacornejo two comments:

1. in our previous LD pruning with PLINK 1.9 there is an option for multi-threading. Then could you clarify what's the motivation to use PLINK2 in this case? 
2. I'm fine with adding more paramters to merge_plink as long as it is backward compatable. That is, you should set default parameters of filters all to zero

dianacornejo:
@hsun3163 will add the plink2 part and we won't make any more changes to it. @gaow I honestly don't remember why using plink2 worked for me and plink1.9 did not work...I'll need think about it. I'm not sure about backward compatibility but I don't see why there should be any issues

PR #478, opened 2022-12-09
Update RNA_calling.ipynb

### LINKED ISSUE #475: Bug fix in RNA calling pipeline
Dear Prof. Wang and other mates:

It has been a long time since I leave this project team. Currently, I am at a new lab and doing some work. Our RNA-seq data is very suitable for me to fully utilize & testing the existing pipeline.

Basically, I am now using the RNA_calling pipeline and have experienced these problems:

* No pipeline is designed for single-ended RNA-seq data in adaptor trimming (using `Trimmomatic`)
  Our data have multiple adaptor sequences in each lane. I tested your recommended tool `fastp` but it did not work very properly. We go back to the  `Trimmomatic` and the following new workflow can be used to deal with SE RNA-seq:
```
[trimmomatic_trim_adaptor_SE]
# Path to the software. Default set to using our rna_quantification.sif image
parameter: trimmomatic_jar = "/opt/Trimmomatic-0.39/trimmomatic-0.39.jar"
# Illumina clip setting
# Path to the reference adaptors
parameter: fasta_with_adapters_etc = path(".")
parameter: seed_mismatches = 2
parameter: palindrome_clip_threshold = 30
parameter: simple_clip_threshold = 10
# sliding window setting
parameter: window_size = 4
parameter: required_quality = 20
# Other settings
parameter: leading = 3
parameter: trailing = 3
parameter: min_len = 50
input: fastq, group_by = is_paired_end + 1, group_with = "sample_id"
output: f'{cwd}/{_sample_id}_trimmed_{_input:bn}.gz'
task: trunk_workers = 1, trunk_size = job_size, walltime = walltime, mem = mem, cores = numThreads
bash: container=container, expand= "${ }", stderr = f'{_output:n}.stderr', stdout = f'{_output:n}.stdout'
    java -jar -Xmx${java_mem} ${trimmomatic_jar} SE -threads ${numThreads} \
        ${_input} \
        ${_output} \
        ILLUMINACLIP:${fasta_with_adapters_etc}:${seed_mismatches}:${palindrome_clip_threshold}:${simple_clip_threshold} \
        LEADING:${leading} TRAILING:${trailing} SLIDINGWINDOW:${window_size}:${required_quality} MINLEN:${min_len}
```
* Small bug in `fastqc`: 
   The line `unzip -o ${_output[0]:n}.zip -d ${cwd}` will leads to an error when using SE data. Though this may not have any impact on the end result.
* Some small bugs are in `STAR_align_1` main workflow:
  * The first is about the line `--sjdbOverhang ${sjdbOverhang if sjdbOverhang != 0 else _read_length} `. In some cases, one may not know the read length in constructing the very first `sample.list`. Only after QC you may have idea of the average read length. In the documentation of `STAR`, it claimed that using the default value `100` will work as well as the optimal value. If the read length column is not provided at first stage, there will raise an error.
  * The second is about the `rm -r ${_output[0]:nnnn}._STARtmp`. It seems that it should be replaced by  `rm -r ${_output[0]:nnnn}_STARpass1` (based on our running result) 
* Another small bug in the `rsem_call_4, rnaseqc_call_4`
  In `R` function `readPicard.alignment_summary_metrics`. To read this file, the line `m <- read.table(files[i], header=TRUE, sep="\t", comment.char="#", stringsAsFactors=FALSE, nrows=2)`, the `nrow = 2` parameter should be corrected to **`nrow = 1`** as the function included the header line. Otherwise, there will raise an error.
  
Thanks a lot for this wonderful pipeline, which literally freed a bunch of my time. Hope all these bug reports will be helpful on your end too!

Gou-29:
Fix #475 in correct folder

### DIFF
--- a/code/molecular_phenotypes/calling/RNA_calling.ipynb
+++ b/code/molecular_phenotypes/calling/RNA_calling.ipynb
@@ -100,6 +100,7 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "id": "rough-robinson",
    "metadata": {
@@ -116,6 +117,7 @@
     "sample_2 samp2_r1.fq.gz samp2_r2.fq.gz fr 150\n",
     "sample_3 samp3_r1.fq.gz samp3_r2.fq.gz strand_missing 75\n",
     "```\n",
+    "**Warning**: For single-ended (SE) data, your input data should exclude the `fq2` column. For other 2 columns, the `strand` column is optional (the `strand` information can be detected in `STAR_output` pipeline) and the `read_length` colunm (can be validated/infered from `fastqc` report) is **required**. If no prior information is avaiable for `read_length`, leave the column empty\n",
     "\n",
     "All the fastq files should be available under specified folder (default assumes the same folder as where the meta-data file is)."
    ]
@@ -195,11 +197,12 @@
     "    --data-dir data \\\n",
     "    --STAR-index reference_data/STAR_Index/ \\\n",
     "    --container containers/rna_quantification.sif \\\n",
-    "    --reference-fasta reference_data/GRCh38_full_analysis_set_plus_decoy_hla.noALT_noHLA_noDecoy.fasta.fasta \\\n",
+    "    --reference-fasta reference_data/GRCh38_full_analysis_set_plus_decoy_hla.noALT_noHLA_noDecoy.fasta \\\n",
     "    --ref-flat reference_data/Homo_sapiens.GRCh38.103.chr.reformated.ERCC.gtf.ref.flat "
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "id": "f04410f9-6721-4bd5-8aea-ef9150103347",
    "metadata": {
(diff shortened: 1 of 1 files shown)


hsun3163:
Hi @Gou-29, thanks for adding more to the pipeline. I wonder, is there a reason u want to keep the STARtmp files? Based on the STAR author these are to be removed and also the STAR pipeline from TensorQTL remove them as well.

```
"    rm -r ${_output[0]:nnnn}._STARtmp\n",
```

Gou-29:
Hi @hsun3163 : in my running on single-ended data, there does not have a _STARtmp file so I removed that line. I do not know whether in PE data there will be one

hsun3163:
> Hi @hsun3163 : in my running on single-ended data, there does not have a _STARtmp file so I removed that line. I do not know whether in PE data there will be one

Per my understanding this is to rm the cache file in case STAR fail to do so themselves, could u push a new commit without it being deleted?

Gou-29:
> > Hi @hsun3163 : in my running on single-ended data, there does not have a _STARtmp file so I removed that line. I do not know whether in PE data there will be one
> 
> Per my understanding this is to rm the cache file in case STAR fail to do so themselves, could u push a new commit without it being deleted?

Actually in my running, if I do not remove that line the code will return an error.

Gou-29:
@hsun3163: I just reproduced the error message:

*in .log file*:
```
rm: cannot remove 'xxx/20Pre._STARtmp': No such file or directory`
```

*in console*:
```
ERROR: STAR_align_1 (id=65bd420a35f37e23) returns an error.
ERROR: [STAR_align_1]: [0]: Executing script in docker returns an error (exitcode=1, stdout=xxx/20Pre.Aligned.sortedByCoord.out.stdout).
The script has been saved to xxx.sh. To reproduce the error please run:
``docker run --rm -xxxx
[STAR_output]: Exits with 4 pending steps (STAR_output, strand_detected_1, strand_detected_2, picard_qc)
```

hsun3163:
> 

Hmmm it is strange, is the tmp file ever generated on your case?

Gou-29:
> @Gou-29
> 
> Hmmm it is strange, is the tmp file ever generated on your case?

Yes, when the pipeline is initially running, there will be three files end with `_STARgenome, _STARpass1, _STARtmp`. As my previous running exclude the line remove  `rm -r xxx_STARtmp`, maybe this file have been remove by STAR itself.

hsun3163:
@Gou-29 I have added various accommodations of the SE in my new PR, incorporating some changes you suggested. I will close this PR now. can you try to pull once my new pr is merged and see if that works? particularly the trim adaptor part

Gou-29:
@hsun3163: Thanks a lot for your updated pipeline! I have tested most parts of it can still, but there are two 3 issues:
* In `fastqc`: I notice that there is an option that the input fastq can be compressed or uncompressed. If my input is `xxx.fq.gz`, the output will be `xxx.fq_fastq.html`, and the pipeline failed to capture this. This small bug will not trigger the end of pipeline and will just throw out an error. 
* In `trimmomatic_trim_adaptor`: since the PE data and CE data have different java configurations, the first line of the actual code should be `java -jar -Xmx${java_mem} ${trimmomatic_jar} ${"PE" if is_paired_end else "SE"} -threads ${numThreads} \`
* In `STAR` part: still, I need to exclude this line `rm -r xxx_STARtmp` to circumvent the error message (which may lead to a direct stop of the pipeline)

PR #489, opened 2022-12-22
force gene annotation to only output gene_id

hsun3163:


gaow:
It seems we are making the output less generic ... why would we want to do that , especially when gene_id is not relevant ot some QTL types?

hsun3163:
> It seems we are making the output less generic ... why would we want to do that , especially when gene_id is not relevant ot some QTL types?

This step is only relevant to the external RNAseq count table (namely the pseudobulk one), which used gene_name instead of gene id. Disallowing the gene_name output is consistent with what we did in the rna calling pipeline.

PR #492, opened 2023-01-03
update replication analysis notebook

jackytao18:
replication analysis notebook

### DIFF
--- a/pipeline/Replication_analysis.ipynb
+++ b/pipeline/Replication_analysis.ipynb
@@ -1,0 +1,177 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "4ceef0cf-b55d-44f6-883e-bef7642aeff1",
+   "metadata": {
+    "kernel": "SoS"
+   },
+   "source": [
+    "# Replication study\n",
+    "This notebook is trying to validate the eQTLs."
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


yuqimiao:
Thanks, Jiajun! The submitted pipeline is currently working with a specific input form, we just created 2 result files with the qvalue and pvalue randomly sampled from a uniform distribution as the input file specified as mwe to run the sos pipeline out; We have the following questions for the generalizability and criteria used in the pipeline.

1. The pipeline uses analysis_1 as the reference and selects the pairs with qvalue < 0.05 as significant; As discussed in the slack discussion, the published/original QTL calling result should be used as the reference since the new analysis is expected to have higher power. However, this setting requires the published results to provide the qvalue or the original pvalue for every feature-SNP pair, is this available for all the molecular phenotypes? If only the latter information is available, we can add 1 step to calculate the qvalues;
2. For the current input form, we require the 2 analysis results to have exactly matched molecular feature names, and the SNP id should follow a chr:position form. Is this matching process generalized enough?

gaow:
Thanks @yuqimiao . All these questions boil down to a simple one -- can we analyze data with these codes (and iron out the issues encountered). Now that @jackytao18 is back on campus we are working on that. We had a meeting already and some plans made. I'll merge it for now but i expect drastic changes in light of the actual analysis.

PR #497, opened 2023-01-16
help multiple row list serve as input

SamuelSun2000:


gaow:
@hsun3163 please review the code above and comment.

hsun3163:
The new mechanism of using the indexes to subset samples in the susie() step will cause a misalignment of x and y. please fix it as I described in slack and resend the PR.

PR #511, opened 2023-01-29
Uncouple unisusie_2 and unwrap list in uni_susie1

SamuelSun2000:
New feature:
1. Now the output rds files are a single susie object instead a list of susie objects.  
**Notice: now the susie object have a extra attribute "name" showing the origin tissue.  and the uni_susie2 utilizes this attribute!** 
Example output file name: demo.ENSG00000197375.unisusie.fit.Ast.rds, demo.ENSG00000197375.unisusie.fit.Exc.rds

2. Now the uni_susie 2 is uncoupled from step1 to be **[susie_output]**, can present the result of uni_susie in a readable way.  Notice: now the input only allow the susie object with attribute "name" added in last step, and the name of gene must be in the file name eg. ENSG00000197375. We can use other strategies if this method is unacceptable. 

The code part is checked to be alright. **Please help me especially check the sos wrapper part, and the ${} variables in R code part.**

The example cmd line will be added after checking this code part. Example input address: /home/hs3393/coloc/susie_output/test/unwrapls/. There are 4 susie objects rds file in this folder generated from updated uni_susie1.


hsun3163:
pls fix the issues described in slack, run the command line until it works, then push again.

PR #514, opened 2023-02-01
revised psichomics normalizaion

Rhopala:
For psichomics part, replace NAs with 0 instead of row means, and get rid of normalization on columns (across events)


gaow:
I htink you need to pad matrices with zero right after the merger. I might be wrong but don't see how this patch can be correct -- this is exactly what I think we should NOT do because this would not distinguish between zero or NA.

Rhopala:
@gaow This is the normalization notebook and it performs the step 4 (convert raw psi values to proper phenotype input for TensorQTL) I mentioned on slack this morning. So the line replacing NA with 0 here is the imputaion right before normaliztion. I'm editing the merger and let it create 0 instead of NA in the `code/molecular_phenotypes/calling/splicing_calling.ipynb` notebook which the end product is the raw psi values table. People will have chance to examine NAs directly produced by Psichomics functions both with the merged psichomics_junctions.txt intermediate output and with the psi_raw_data.tsv final output in the splicing_calling.ipynb notebook.

I'm pushing this normalization notebook first since it would be easier for Minghui to get the revised code and continue on normalization step. He doesn't have the junction merging issue since his data was under the same directory.

gaow:
>  So the line replacing NA with 0 here is the imputaion right before normaliztion.

Then you are doing a zero imputation which is worse than a mean imputation ... please use mean imputation as a place holder for now.

gaow:
It reads reasonable .. I still don't like it that we replace NA with zero after full join, rather than setting some parameter in the full join function to pad with zero by default. However this is not  somethign we can change . I did some google search myself and everybody seems to be doing what you are doing. We will live witi it for now. As you explained the first processing step for the junction files should not create any NA , and we should be safe.

Rhopala:
Yes I did take some time to search if there exist table joining function with option of what to fill first but failed to find any. So just turn to do it in this way ...

PR #515, opened 2023-02-01
Corrected TPM/Count headers; Converted outlier messages to for loops for easy readability

m-mews:


gaow:
Nice @m-mews thanks!!

PR #519, opened 2023-02-06
minor modification for conversion from bam to fastq section

Chunmingl:
Modification is for overcoming the directory restrictions for generating temporary files while running the commands.


hsun3163:
An improvement is included in the PR #521 , this PR is thus closed.

Issue #527, opened 2023-02-15
Remaining problem in unisusie pipeline


SamuelSun2000:
@hsun3163 Please note that `uni_susie_2` in https://github.com/cumc/xqtl-pipeline/blob/main/code/fine_mapping/SuSiE/SuSiE.ipynb cannot be successfully executed yet. There are two problems:

1. Some of the rds files will be blank, and cannot be handled by readRDS() function.  Adding `improt os; skip_if(os.path.getsize(_input) == 0)` may help.
2. Because now output rds is only one susie_object, there will be errors about the variable `susie_list` in `uni_susie_2`. We no longer can use `susie_list[[i]]$pip` to extract pip. Instead, we should directly use `susie_list$pip`.

Because I currently don't know the strategy for correcting this step -- whether I should make a vcf for per gene per tissue, or integrate the result per gene in all tissues.  So I leave this issue here as a reminder.

gaow:
@SamuelSun2000 could you provide me a minimal working example for this? (remember the style of the RNA-seq protocol you have tried last year we have the MWE). I'm swamped today until late in the PM .. but this one is now on the top of my TODO to finalize so others can use.

gaow:
I see. @SamuelSun2000 the MWE you provided me can they reproduce these issues?

SamuelSun2000:
Yes, both of them.

hsun3163:
I was thinking, perhaps we should take away the uni_susie_2/3 altogether. And make the transform to vcf step another utility in the susie_post_proc notebook. Along with the code that produces the three summary tables.

gaow:
yes i remember your proposal. let me take a look maybe this evening or tmr

PR #530, opened 2023-02-16
Currently make the following unisusie steps optional

SamuelSun2000:
To avoid the job exit with errors, I currently muted the unisusie_2 and 3 by suggestion, it will not influence unisusie_1.


gaow:
the patch does not look right ... you cannot make steps optional by just commenting them out -- that's what you did right ? (and without justificcations)

SamuelSun2000:
> 1. why mask? what are you trying to achieve? 2. would we ever need to unmask them? 3. if 1 is justifiable and 2 is necessary, what would be the best way to do that?


1. Now if we run unisusie, it will return an error because unisusie_2 has issues. I blocked it so that people will not get confused, because unisusie_1 successfully yields output without error.  Blocking currently is what @hsun3163 suggests to me.

2. I can see the function is to generate some vcf files. Hao knows the detail  @hsun3163 could you plz explain that?

3. I'm providing two strategies at https://github.com/cumc/xqtl-pipeline/issues/527

gaow:
okay let me merge this then edit .. .

Issue #525, opened 2023-02-14
Different format of snp for Knight(WashU) genotype data


zq2209:
When running `cis` for Knight data, I am getting the error.

>Traceback (most recent call last):
  File "/home/zq2209/.sos/e0257c72c6594a84/singularity_run_4940.py", line 61, in <module>
    pairs_df = pairs_df.assign(
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/frame.py", line 3827, in assign
    data[k] = com.apply_if_callable(v, data)
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/common.py", line 329, in apply_if_callable
    return maybe_callable(obj, **kwargs)
  File "/home/zq2209/.sos/e0257c72c6594a84/singularity_run_4940.py", line 63, in <lambda>
    ref = lambda dataframe: dataframe['variant_id'].map(lambda variant_id:variant_id.split("_")[-2])).assign(
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/series.py", line 3879, in map
    new_values = super()._map_values(arg, na_action=na_action)
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/base.py", line 937, in _map_values
    new_values = map_f(values, mapper)
  File "pandas/_libs/lib.pyx", line 2467, in pandas._libs.lib.map_infer
  File "/home/zq2209/.sos/e0257c72c6594a84/singularity_run_4940.py", line 63, in <lambda>
    ref = lambda dataframe: dataframe['variant_id'].map(lambda variant_id:variant_id.split("_")[-2])).assign(
IndexError: list index out of range

After running line by line, this is because `snp` column in Knight genotype file has a different format compared to ROSMAP genotype data. 

In ROSMAP, the format of `snp` is like`chr1:248945797_G_C`, but in Knight, it is `chr1:732994:G:A`. Knight use `:` instead of `_`. Details shown in attached screenshots. This cause the pipeline return that index error.
<img width="659" alt="Screen Shot 2023-02-13 at 9 31 09 PM" src="https://user-images.githubusercontent.com/112965889/218623843-836181e9-5764-4df7-b814-78ef51d6d66e.png">
<img width="598" alt="Screen Shot 2023-02-13 at 9 31 44 PM" src="https://user-images.githubusercontent.com/112965889/218623844-d1a7a80a-4f66-4028-98bf-02d5ff9744d6.png">

Two possible solutions:
[...]

PR #534, opened 2023-02-20
update king for accomodating both plink1 and 2 FID

hsun3163:
1. Change king2 such that the discrepancy of FID in  #525 should not longer be an issue.
2. Add pos info in susif module


gaow:
> `dat = dat[!duplicated(dat),]`

I wonder if it is logically correct / elegant to have to clean up duplicates like this ... looks like we should have not introduced duplicates in the first place but let me know if you still believe here we know exactly what will be removed rather than ad hoc data clean up ...

hsun3163:
The remove duplicated is a extension from the original function where sorted and remove any Non-uniq IID. The reason that was performed previously is to to remove the the case where
A X B  related
B X C related
So we will have A B B C as related individual to be removed.

This is also indicated in the code body now.

PR #539, opened 2023-02-27
simulation_pipeline

SamuelSun2000:
I created the simulation pipeline, the pipeline works well, but I want to discuss some details here. So please note: this is definitely not the final version.

In the prototype, we randomly generate one colocalized causal variant $X_c$ and then randomly select the other two variants $X_p$ and $X_e$ such that $LD \approx r \pm 0.01$. However, there can be variants that are far away but have a high correlation, even if they are not linked. I will later check on what occasions this can happen.

Should we switch our logic -- first choose a block that has a high average LD, then we can define it to be a high LD block, then find $X_p$ in this block, rather than "randomly" select $X_p$ that satisfies $LD \approx r \pm 0.01$. 

I come up with this question when analyzing the simulation result and find some abnormal results, see another pull request for details.

### DIFF
--- a/code/fine_mapping/Simulation/Simulation.ipynb
+++ b/code/fine_mapping/Simulation/Simulation.ipynb
@@ -1,0 +1,293 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "ae8a823b-fa13-41fb-ba13-ecf52237e4a5",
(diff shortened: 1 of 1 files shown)


gaow:
@SamuelSun2000 thanks and again, let's have this on Xuewei's github repo ...

PR #541, opened 2023-02-27
fix RLE sample outlier error

grennfp:


gaow:
thank you @grennfp I wish we caught it earlier!

PR #542, opened 2023-03-01
mvf SuSiE WIP

hsun3163:


hsun3163:
Needs to figure out a clever way to specify the y_fs and their position.

PR #555, opened 2023-03-17
update mashr and rds_to_vcf

rfeng2023:
Only the necessary parameters were changed, and nothing that produced fake ouput was added


hsun3163:
None of the changes drastically change the pipeline, but simply fix a few things regarding the containers.

PR #557, opened 2023-03-24
accomodation of proteiomics

hsun3163:
Many of the problem we observed today is due to the non-standard pre-processing of the proteomics data. Now the proper support for proteomics are added. The gene_annotation notebook, annot_coord_protein step shall generate a bed file with #chr start end ID and sample names as columns, just like all other bed.gz files, with following features:

1. The #chr will have "chr" prefix to the number
2. The ID is in the form of chr_gene.id_uniportID


hsun3163:
RE the FIXME on susie. It is near impossible to unify the susie complete data interface with the susie RSS interface due to the drastic difference in the input they require and the way tasks are parallelized.

PR #566, opened 2023-04-12
annotation of snp temp module added

hsun3163:


hsun3163:
For susie_sqtl, Switch back to 
```
      read_gene_pheno = function(path){
      arg = paste0(c("grep  ${_genoFile["ID"]}  ",path ),collapse = "")
      result = system(arg,intern = T)
```
So that the extraction of phenotype is independent of pos.

PR #601, opened 2023-07-04
update sample name conversion in sQTL

rfeng2023:


hsun3163:
The sQTL update is mainly to accommodate the updated pandas version, where "squeeze=True" option no longer available.

PR #602, opened 2023-07-10
Update build_container.ipynb

Jiacheng0007:
Hi, the updated notebook was used to build the containers with a default def file and use the parameter `entrypoint` in the newest version of sos (0.24.1) to use the containers.


Jiacheng0007:
I will open a new pr

PR #634, opened 2023-08-30
update padding

rfeng2023:
Hi @gaow We also need to add output `mu` to the padding pipeline, please check if my changes are correct. If it's correct, I'll combine all the outputs into a list that will directly interface with our intact pipeline


rfeng2023:
Hi @gaow , I've changed the way to generate mu and coef as we talked in the meeting

PR #648, opened 2023-09-16
susie pipeline fix

hsun3163:


gaow:
thanks @hsun3163 at the face value it looks it is completed? I need to review the logic in the utils script a bit more carefully but it looks great! For  this line and comment:

>     for (r in 1:length(fdat$residual_Y_scaled)) { ## Cant have a universal way to specify names due to the accomodation of missingness, use index instead

Can we use `residual_X_scaled` instead because we will always have it? I think it is important to keep the actual name of `r` in the RDS file that we save.

hsun3163:
> thanks @hsun3163 at the face value it looks it is completed? I need to review the logic in the utils script a bit more carefully but it looks great! For  this line and comment:
> 
> 
> 
> >     for (r in 1:length(fdat$residual_Y_scaled)) { ## Cant have a universal way to specify names due to the accomodation of missingness, use index instead
> 
> 
> 
> Can we use `residual_X_scaled` instead because we will always have it? I think it is important to keep the actual name of `r` in the RDS file that we save. 

I have tested the current pipeline last night and it runs while specifying two phenoFile and covFile, I.e two condition. I use residY b/c, when we want to run mvsusie, there will be only 1 y, and multiple residX, but we only want to run mvsusie once.

I feel like it is quite hard to specify the name of r when you need to correctly drop the name while a phenoFile is missing.

gaow:
> I feel like it is quite hard to specify the name of r when you need to correctly drop the name while a phenoFile is missing.

That is true. But that is only relevant to mvSuSiE where there will not be this line anyways: `r in 1:length(fdat$residual_Y_scaled)` right? So what's in SuSiE related code should work well with `names`

PR #679, opened 2023-09-20
LMM environment

dianacornejo:
Added packages for the automatic built of sif image for LMM environment


danielnachun:
Thanks for starting this PR @dianacornejo.  In order for the GitHub Actions pipeline to rerun, you'll have to reopen this PR from a direct copy of this repo and not your fork.

Some other fixes that will be needed:
1) All of the packages must already in exist as `conda` packages, but it looks like a few of them might not?
2) The package names must be written exactly as they are spelled as the `conda` package.  You can get the exact name from the URL of the `conda` package
3) All of the fields except notes need to be filled in.  The only exception is that universal packages are not versioned.  

I will download this PR and start filling in the missing information and push it the changes as a new commit to your new PR for this.  Thanks for being the first tester of the pipeline!

danielnachun:
Also I see that you put the changes into the `misc/xQTL_conda_packages.csv` but the file we are using now is `container/containers.csv`.  That has all the proper formatting that we need and it should be clearer how to add the packages.

dianacornejo:
@danielnachun I'm sorry I did not know about the other file `container/containers.csv`. I'll just add the specific rows for my enviroment in that file. There are 2 software that I could not find conda packages for it GMMAT and hudson (this is R specific for the creation of hudson plots). Also, I do have to fill out the version number. I saw some of the other rows did not have it so I wasn't sure. Thanks for all the comments. I'll address them and ask for another PR.

PR #684, opened 2023-09-25
in progress documentation updates to RNA-seq notebooks

grennfp:


gaow:
@grennfp I see you tried out doing transQTL per chrom. I don't recall discussing that. Did you get your analysis to work this way? I'm actually curious. There are more consideration to this than simply changing it to per chrom. I also have some more questions RE this PR ... perhaps it is more productive if we talk on zoom? I'll DM you about it.

PR #685, opened 2023-09-25
TensorQTL Trans per chrom rough draft

hsun3163:
Same interface as the cis module. Should be able to run with the same commands as the cis just change the step name from cis to trans, untested. 

No permutation applied.

trans_arch use the same interface as the old trans analysis, but remove the part with permutation. I feel like trans_arch should be able to run with reasonable mem. Also untested.


hsun3163:
I suppose the trans per chrom is implemented in another version so this PR is safe to be closed?

gaow:
@hsun3163 there is still something wrong with the trans-QTL analysis that Frank is taking a look at. I'm closing this for now and will adopt Frank's finalized version later.

PR #691, opened 2023-10-02
update rna_quantification container

rfeng2023:


danielnachun:
It looks like this worked!  Can you add the version for `deeptools`?  We can use the latest version 3.5.4.  And also please use all lowercase (`deeptools` instead of `deepTools`).  Conda is not case sensitive but I'd like to keep to all lower case for consistency.

danielnachun:
Can you also add `r-ape` 5.7 to the `rna_quantification` container?  We discovered that it seems to be missing.

PR #709, opened 2023-10-10
fix susie

rfeng2023:


gaow:
Thanks @rfeng2023! A few comments on writing neat codes:

1. Don't put any interactive analysis or example commands after `global` section when you already started the pipeline implementation. These should be in the example section above the implementations. 
2. Please always use relative path not absolute path in your example, to make it easy for people to copy-paste to test it out
3. We should run univariate regression only on variant_index selected, not on all variants up-front. 

I'm going to try fixing 1-3 now and let u know. For 1-2 please keep in mind so we avoid that next time.

rfeng2023:
@gaow Got it, thanks!

PR #714, opened 2023-10-11
robustness improvement

hsun3163:
1. Add explict naming for susie
2. add robustness for gene_annot: now it will match the colnames with a list of known potential colnames that could ends up in a bed like file, and drop those overlap cols.


gaow:
@hsun3163 sorry we have made lots of changes already to these pipelines -- I did not ask for your input explicitly assuming you're busy with other things. Next time I will move these discussions to xqtl-protocol channel. Short story is that 1) I took care of fixes in SuSiE.ipynb and @grennfp did TensorQTL.ipynb. Your commit now will overwrite what we did. 

I think we need to study your patch a bit more to decide what to adopt. Let's talk on slack about it.

PR #715, opened 2023-10-11
fix hardcode

rfeng2023:
Fix the hardcode in two finished mash pipeline. Would clean up the rest ones.


gaow:
Thanks @rfeng2023 i'll merge this. Let's also discuss on slack to see how we prioritize the rest.

PR #723, opened 2023-10-16
gene annot updates

hsun3163:
the purpose of this module is to handle count table that are not produced by us and provides an alternative entry point.
What this function does, in the old version, is it see if any element in the given list [“chr”,”start”,”end”] are present in the column names of a df. If so, then drop all three columns.
This implementation is problematic, for it will drop columns that are potentially not presented in the df.
So in my new implementation, I drop the column based on intersection of the df actual column and the provided list.
 (edited) 
[3:56](https://wanggroup.slack.com/archives/D02RDSPKSP5/p1697054177428459)
This is the reason pertaining to changes 1
[3:59](https://wanggroup.slack.com/archives/D02RDSPKSP5/p1697054348519739)
Given the updated mechanism in changes 1, this function now remove any columns that we don’t like based on a given list, so I extend that list to include many names I remembered seeing that could means chr start and end. Such that this module can truly do what it design to do, take in some random bed.gz or count table and then re annotate them


gaow:
Thanks @hsun3163 but there's a  merge conflict. Could you please resolve it?

PR #734, opened 2023-10-25
add sos in stephenslab image

rfeng2023:


gaow:
Hmm I'm not sure if we want to do it -- if so we would have to do this for all our images. Let's discuss with MemVerge see what would be the best way moving forward. Essentially what we need is to be able to run the `stephenslab` image inside of another separate `sos` image. I'll close the PR for now.

PR #739, opened 2023-10-30
data preprocessing updates + example files

alexmccreight:


gaow:
@alexmccreight I'll merge the PR, even though that in PCA you removed the text on dealing with related individuals which should not be removed. We can always dig that up from tag 0.1.0 as the next step. Currently the correct formatting and improvement of documentations for other notebooks are more important.

PR #771, opened 2023-12-18
fix the formatting of input filenames

RuiDongDR:
Fix the formatting of input filenames
- global_parameter
- add_chr step: use "genoFile" instead of filename
- qc_3 step: write the output to "${_output[0]:n}.stdout\n"


gaow:
I'll close this PR because as discusesd the first two changes need further work. The last change was valid and @hsun3163 adopted it in his other PR.

PR #782, opened 2023-12-23
Update MMQTL, gene annotation and tensorQTL for missingness

hsun3163:


hsun3163:
Changes:
1. MMQTL : change from group_by_region to for_each
2. TensorQTL, update missingness handling mechanism in cis, will update trans after this have been successful for a while for cis
3. GWAS QC, for genotype data with only unrelated samples, now it will still produce the unrelated.bed so that the qc for unrelated sample does not need to be done again.
4. gene annotation, add handling of  A|gene_id pattern to accommodate the gtprotein by hans group
5. Phenotype_formatting
5.1 Phenotype_per_chrom: a new region list documenting what file after partitioned will contains each region are generated, to accommodate huge phenotype file like atac seq or methylation
5.2 phenotype_per_chrom_gct: instead of a bed.gz file, this will create a gct matrix for the partitioned phenotype file, for mmQTL.

gaow:
Thanks @hsun3163 at this point I need to be super careful about any changes made. for the issues i spot:

1. instead of 
```
open(_output):
  pass
```
which basically creates empty output, can we move the `done_if` statement to before the input/output, right after related_id is checked?

2. What's the reason you use a separate step to output related individuals? I prefer it in the same step where if there are related individuals to output, we output it as is, otherwise, we touch the output empty file. I think it can be achived by some bash script combined with SoS syntax:
```
if [ ${some python codes to generate the criteria} eq 0]; then
   ...
else
   touch ${_output[1]}
fi 
``` 
3. The way you handle missing data in tensorQTL, `phenotype_df = phenotype_df.dropna(axis=1)` is it going to drop the entire sample if it is missing in ANY of the molecular phenotypes in this matrix? It is not the desired behavior, if that is the case. We should only drop the missing values per molecular phenotype. Is that doable within tensorQTL framework?

gaow:
Also, at the end of this PR: https://github.com/cumc/xqtl-pipeline/pull/771/files there is a small fix that warrants incorporating. Otherwise the other fixes proposed in this PR i'll not accept them. Please incorporate this small fix to your PR so I will close the other PR.

hsun3163:
> Also, at the end of this PR: https://github.com/cumc/xqtl-pipeline/pull/771/files there is a small fix that warrants incorporating. Otherwise the other fixes proposed in this PR i'll not accept them. Please incorporate this small fix to your PR so I will close the other PR.

That is, incoperating only `qc_3 step: write the output to "${_output[0]:n}.stdout\n"` ?

hsun3163:
> Thanks @hsun3163 at this point I need to be super careful about any changes made. for the issues i spot:
> 
> 1. instead of
> 
> ```
> open(_output):
>   pass
> ```
> 
> which basically creates empty output, can we move the `done_if` statement to before the input/output, right after related_id is checked?
> 
> 2. What's the reason you use a separate step to output related individuals? I prefer it in the same step where if there are related individuals to output, we output it as is, otherwise, we touch the output empty file. I think it can be achived by some bash script combined with SoS syntax:
> 
> ```
> if [ ${some python codes to generate the criteria} eq 0]; then
>    ...
> else
>    touch ${_output[1]}
> fi 
> ```
> 
> 3. The way you handle missing data in tensorQTL, `phenotype_df = phenotype_df.dropna(axis=1)` is it going to drop the entire sample if it is missing in ANY of the molecular phenotypes in this matrix? It is not the desired behavior, if that is the case. We should only drop the missing values per molecular phenotype. Is that doable within tensorQTL framework?

For 3, not unless we change the tensorQTL embedded functions, for the phenotypes are input as a matrixs. I am not sure how NA in phenotype will impact the results though. Perhaps tensorQTL omit them automatically so there is no need to fill in the NA.  Could be tested.

For 1,2 let me try to change it into that way. For 2 I was trying to keep the bash part simple.

gaow:
@hsun3163 thanks. Let's revoke change for tensorQTL,  but instead raise an issue on their github repo for this feature request. I will follow up with them if they dont respond. 

Also for phentype by chrom, can we consolidate these two files into one? It seems redundant to have two files.

hsun3163:
> @hsun3163 thanks. Let's revoke change for tensorQTL, but instead raise an issue on their github repo for this feature request. I will follow up with them if they dont respond.
> 
> Also for phentype by chrom, can we consolidate these two files into one? It seems redundant to have two files.
Posted, but there are other posts asking about NA in phenotype data and I don't think they are too keen on handling this issue but suggest the user to impute.
https://github.com/broadinstitute/tensorqtl/issues/117

gaow:
@hsun3163 for tensorQTL pipeline you removed these lines:

```
    covariates_df = pd.read_csv(covariates_file, sep='\t', index_col=0).T
    pr = genotypeio.PlinkReader(plink_prefix_path)
    genotype_df = pr.load_genotypes()
    variant_df = pr.bim.set_index('snp')[['chrom', 'pos','a0','a1']]
    ## Retaining only common samples
```

are you sure about it?

hsun3163:
Fixed

PR #797, opened 2024-01-16
update fsusie in cis workhorse

hsun3163:
1. add temp code to overwrite function in pecotmr (to be removed after container updates)
2. add a module to check the number of phenotype marker in a tadb for a phenotype file, and create the region_list index for that phenotype file for all the tadb that satisfy the criteria (to mimic what we did in the fsusie manuscript)

2 is tested. 1 is tested up to the point of actually completing fsusie before minerva keep giveming me core dump. The test data is using atac seq data.


hsun3163:
The command spliter update is also here. The test file used for command spliter are attached.
[test_command_region_list.txt](https://github.com/cumc/xqtl-pipeline/files/13984695/test_command_region_list.txt)
[test_sos_command.txt](https://github.com/cumc/xqtl-pipeline/files/13984696/test_sos_command.txt)
[test_aws_command.txt](https://github.com/cumc/xqtl-pipeline/files/13984697/test_aws_command.txt)

PR #798, opened 2024-01-17
test on mrmash_wrapper with cis_analysis pipeline 

Chunmingl:


Chunmingl:
The test is done, and I have PR’d the edits for the workflow, I made some edits on the misc.R  but I did not want to make too much of a change on functions since it will be used in other pipelines.

This update only works when the pecotmr is updated to the current github repo source code, I loaded the package `devtools::load_all("~/githubrepo/pecotmr")` before running the command of `load_regional_multivariate_data`, the devtools::load_all step is removed from the PR version.

For testing purposes, I also modified the the input X snp numbers, `res = mrmash_wrapper(X=fdat$X[, 1:150], ...` and it is also removed from PR version. Without modification, there will be 229085 snps that need to be processed.

Example command is added as well.

PR #801, opened 2024-01-21
RSS_1 raise imputation

SamuelSun2000:
Remaining problem: now maybe cannot simply library(susieR), should use the previous version?


SamuelSun2000:
We need to make sure the susieR version using will get the same result...

gaow:
@SamuelSun2000 as i mentioned before please send PR to gaow/susieR to undo the changes I made before. So we finalize on that version of susieR.

gaow:
@SamuelSun2000 for TRUE and FALSE parameters, please see an example here: 

https://cumc.github.io/xqtl-pipeline/code/cis_analysis/cis_workhorse.html#univariate-analysis-susie-and-twas

`twas_weights = True` this is how we are supposed to use Boolean parameters. Please change your code to this convention

PR #804, opened 2024-01-22
leafcutter normalization fix

grennfp:


gaow:
@grennfp fillna has an option to fill it in place: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html see inplace = True. I think it is a clearer alternative?

PR #806, opened 2024-01-23
pseudobulk processing

hsun3163:


hsun3163:
Didnt realized this was not added. There are two version of QC documented in this notebook. 1 is based on Nick's script and the published paper. 2, `SE_qc` is based on Panos' lab's R package variantpartition. 2 is based on the singlecell experiment data object and I don't think it suit out needs.

PR #830, opened 2024-02-05
randomize region

hsun3163:
tested


gaow:
I've already did it (and mentioned it on slack but maybe you missed it): https://github.com/cumc/xqtl-pipeline/commit/c45678585a57590ddca3c4c7c63db61c2c994a72 Shall we close this PR ? I think it's okay either to shuffle command or region, right?

PR #839, opened 2024-02-08
Fixed and error in gene_annotation.ipynb and created new notebook for psuedo bulk eqtl QC and normalization

al4225:
1. Fixed an error in phenotype preprocessing: gene_annotation.ipynb annotate_coord_gene, based on the sample-participant-lookup table. The sample-participant-lookup table path should be: /mnt/mfs/ctcn/datasets/rosmap/wgs/ampad/qualityControl/sampleSheetAfterQc.csv

2. The notebook pseudobulk_pheno_aggregation.ipynb includes sos workflow of 3 versions of psuedo bulk eqtl seurat rds objects, from aggregation to normalization. Comments and details are in the notebook.


gaow:
@al4225 for your gene_annotation notebook my issue is that you should make it backwards compatible. `,` instead of `\t` looks like it's going to break things. But your changes itself looks good. Just fix the separator it should work?

rfeng2023:
Hi @gaow I have some questions here
1. we don't have write access to this file, somehow it would report error like

> sample_participant_lookup
> PosixPath('/mnt/mfs/ctcn/datasets/rosmap/wgs/ampad/qualityControl/sampleSheetAfterQc.csv')
> sample_participant_lookup.is_file()
> False

2. That means,  this function actually not work

>     def rename_samples_using_lookup(df, lookup_path):
>         sample_participant_lookup = Path(lookup_path)
>         if sample_participant_lookup.is_file():
>             sample_participant_lookup_s = pd.read_csv(sample_participant_lookup, sep="\t", index_col=1, dtype={0:str,1:str})
>             df.rename(columns=sample_participant_lookup_s.to_dict()["genotype_id"], inplace=True)
>         return df


3. Even if that work, this file is separated by `,` instead of '\t', it should report error
4. And even if I used `,` here , there is no column named 'genotype_id'

> ProjID,SampleID,BloodBrain,Tissue,ExtractionKit,QualityCheck,Reason,Note,VarsCalled,GenotypePc,PC1,PC2,PC3,PC4,PC5,PC6,PC7,PC8,PC9,PC10,PF_READS,PF_READS_ALIGNED,PCT_PF_READS_ALIGNED,PERCENT_DUPLICATION,ESTIMATED_LIBRARY_SIZE,MEDIAN_COVERAGE,MAD_COVERAGE,PCT_EXC_DUPE,PCT_EXC_TOTAL,MEDIAN_INSERT_SIZE,MODE_INSERT_SIZE,MEDIAN_ABSOLUTE_DEVIATION,AT_DROPOUT,GC_DROPOUT

Then why we are changing our initial lookup table (two columns separated by '\t') to this file here?

gaow:
@rfeng2023 

> we don't have write access to this file, somehow it would report error like

interesting. I did not know path.is_file() has this behavior ... then you can try change it to `os.path.exists()` something like that using the more traditional `os` package, or maybe you fixed that already in your PR?

> this file is separated by , instead of \t, it should report error

can we support both using regex, eg, `df = pd.read_csv('file.csv', sep='[,\t]')` something like this? 

> And even if I used , here , there is no column named 'genotype_id'

If we are sure that for a look-up table, the first and 2nd column should be genotype_id and phenotype_id, then we can safely use the loaded file (with header) regardless of the column names? If that is the case, then would it work if we load the example file that you have under 4, via `df = pd.read_csv('file.csv', sep='[,\t]')` with header, and use first two columns as the look up table?

PR #840, opened 2024-02-09
update susie post processing

rfeng2023:


gaow:
Thanks @rfeng2023 Apparently some cleaning up is needed but it looks decent. Merging it for now

PR #841, opened 2024-02-12
add sumstats export

rfeng2023:


rfeng2023:
@gaow  I added the sumstats as a separate export object, will run all susie results with that pipeline if you think is fine. 
In addition, `genoFile` and `phenoFile` etc. are essential in global of cis_workhorse, which are not needed in post processing. should I move them into the separate cell such as susie_twas?

rfeng2023:
Hi @gaow , I've added meta dictionary for post processing, but in such case, we won't be able to run sQTL simultaneously, as the raw name for eQTL and sQTL in the meta are same, and the output would be a table with eQTL conditions only.

gaow:
@rfeng2023 i will talk to you a bit more later today.

PR #853, opened 2024-02-17
Susie_rss update

SamuelSun2000:
1. Take out loading region outside loop so no need to load LD twice.
2. I prefer to make the study name appear on file name so that I can directly know what studies I've already analysed. Instead of putting them all into a large rds output.


gaow:
@SamuelSun2000 sorry we still want all studies for the same region going into the same large file. This is for convenience of data integration and comparison purposes. We have an extra later step to address your concerns regarding 'transparency'. For now please revert it back to everything into one RDS file. Thanks!

PR #864, opened 2024-02-21
first version for gwas post and overlapping with qtl

rfeng2023:
some issues
1. gwas_results_export_2 is extracting union cs and pip of all datasets and methods. The file size of ouput is 608K for 2 datasets and 7 methods. I think that is acceptable for 8 datasets. 
2. overlapping between qtl and gwas is assuming there is no flip issue, and just intersect with `chr:pos:ref:alt` , as I believe there are different mutation could occur on same position.


gaow:
> overlapping between qtl and gwas is assuming there is no flip issue, and just intersect with chr:pos:ref:alt , as I believe there are different mutation could occur on same position.

Great point ... no at this point we don't assume that. In TWAS we flip on the fly to make xQTL and GWAS data compatible. We do the same in enrichment analysis. Have to think of what we need to do if anything. 

Glad you are starting to even address the overlapping question. We can discuss more tomorrow see what is the best way but i think we should really just try a few things to first learn the data better, and decide what's most reasonable

PR #868, opened 2024-02-22
Generate sample list & remove region list restriction

wangyf1125:
Generate sample list in`.txt` file & remove region list restriction in trans analysis.


gaow:
@wangyf1125 why would you need to add a step to output sample list?

wangyf1125:
@gaow 
We might have this conversation before:
> Yifei Wang: I am wondering if we need to keep a sample list as I need to construct E and G, and I want to keep the samples same with sumstats.  If so, should I add this part in TensorQTL.ipynb or my own pipeline?

> gaow: good point. In fine-mapping pipeline we do keep a list of samples that we actually analyze. I'm not sure how easy it is to add to tensorQTL pipeline but ideally yes it should go there.

So I suppose I could add a step to output sample list here?

PR #873, opened 2024-02-26
gene_annotation:convert ProjID to SampleID

al4225:
convert first column to second column: convert ProjID to SampleID


gaow:
Still, the previouls code assumes `\t` but your updated is `,`. Why don't we stick to `\t`? This seems to be breaking the previous anlaysis code if I merge the change?

al4225:
> Still, the previouls code assumes `\t` but your updated is `,`. Why don't we stick to `\t`? This seems to be breaking the previous anlaysis code if I merge the change?

Cuz the reference data sampleSheetAfterQc.csv is a csv file and uses ','. If we stick to \t, it will return an error.

gaow:
then we should make that an option. parameter input `comma` or `tab` ?

al4225:
yes thanks I can do that.

________________________________
发件人: gaow ***@***.***>
发送时间: 2024年2月26日 17:07
收件人: cumc/xqtl-pipeline ***@***.***>
抄送: Liu, Anjing ***@***.***>; Author ***@***.***>
主题: [EXTERNAL] Re: [cumc/xqtl-pipeline] gene_annotation:convert ProjID to SampleID (PR #873)


then we should make that an option. parameter input comma or tab ?

―

al4225:
Already modified code considering **\t** and **,**.

PR #874, opened 2024-02-27
update AD post processing 

rfeng2023:
using `lapply` instead of for loop here, as I think I am loading exported qtl file anyways, it should not cost too much memory, please let me know if that is not acceptable @gaow


gaow:
@rfeng2023 thanks! It's a bit hard for me to digest it at this point ... (i even stared at the title of this PR "Adpost" for quite a while trying to figure out what that word means).  I guess as long as we can verify the outputs are correct, we are more or less reassured. One thing to caution is that next release of top_loci table will be NULL if there are no rows. So `has_rows()` related function will be broken. We can talk tmr.

PR #875, opened 2024-02-27
Adjust output file name

wangyf1125:
Adjust output file name to phenotype basename


wangyf1125:
@gaow Originally, the output file names are from phenotype basename. I suppose phenotype might be more representative among datasets but I don' know.

gaow:
Thanks @wangyf1125 this is a good patch.

PR #886, opened 2024-03-02
fsusie fix (test) 

hsun3163:
fsusie:
1.Not removed fsusie trimmed
2.Set colnames of residual X as names for pip, which should be translated to the variants name in the final fsusie object.
3.add the Y coord for each epi marks and peter reminded.

Following are not added:

rss
1.extract gene name when available from region list
2.extract only the rows in a eQTL dataset that has the gene name
3. pick top 1 signal from the gene name
4. when flagged, compute LD from genotype


gaow:
@hsun3163 I don't think we should create a separate workflow for eQTL.  We should use the same RSS workflow for everything unless there is a very good reason to split it, which you need to argue about. Also please don't delete the workflow that makes slides. They are broken now but they can be fixed up for use in the future.

hsun3163:
The slides steps are broken to the point where nothing could be dryrun even with the step title commented out so I moved them in another notebook.

The distinction of eQTL and gwas is that the eQTL have one snps corresponding to multiple genes and require the filteration for each. I think if we put the read eQTL function in the pecotmr it could be simplify using a if statement and combined later.

gaow:
> The distinction of eQTL and gwas is that the eQTL have one snps corresponding to multiple genes and require the filteration for each. 

I see, looks like the load summary stats function in pecotmr needs to be changed with an additional parameter like "trait_ID" whcih is set to NULL by default, and if it is not NULL we use the gene name to filter? 

I would rather implementing that right now and finalize on the RSS analysis notebook than to combine it "later" which can be indefinite.

PR #905, opened 2024-03-08
accomodate s3 path

hsun3163:
Counterintuitively, the output path of aws runs is effectively empty. So additional parameter are needed to account for the differences.


gaow:
@hsun3163 thanks ! this was counter intuitive to me in the beginning but if you think of github files vs folders it's the same thing: there is no such a concept as folders on github or AWS. Its always about files. With that in mind perhaps you see it starts to make sense (hopefully you get what i mean)

PR #919, opened 2024-03-15
add more tryCatch in export

rfeng2023:


gaow:
@rfeng2023 typically for tryCatch usuallly a warning is made before falling back to a trivial behavior, unless in this case you are 100% sure that users need not be warned.

rfeng2023:
> if (${\"TRUE\" if exported_file.is_file() else \"FALSE\"}){",
>     res_exp <- tryCatch({readRDS(combine_data)}, error=function(x){return(NULL)})

Hi @gaow The logic implemented in this command is designed to handle the processing of exported meta files from previous scenarios. When an exported file is supplied, the command verifies the presence of an existing exported file with specified data (such as geneA with ROSMAP). If found, and there's a need to append new data (like geneA with MSBB), the command reads and merges the new data with the existing file. In contrast, if it's necessary to add data for a different entity (like geneB) for which no exported file exists (unlike geneA, which was previously exported), the command would not report error by read geneB file.  This allows for flexible handling of adding new conditions or incorporating additional genes into the export process.
The choice to omit a warning in the second scenario stems from the belief that the readRDS function is unlikely to encounter numerous failure scenarios, without warning would avoid filling the log file with extensive warnings during the export of multiple genes. However, if adding a warning is considered more prudent for enhancing the process's clarity or for tracking purposes, I am open to incorporating it.

gaow:
Then sounds like a clearer logic is that instead of tryCatch, an if else statement shoudl be used?

The general point is that tryCatch is typically designed to ensure that most of the case it works out and fail only at unexpected cases. For example to read a file, something like:

```python
try:
   read_file(file)
except:
   warning_message()
   pass
```

is less clear than

```python
if file.exists():
  read_file(file)
```

That said, I'm neutral about whatever you do, as long as you can comment it a couple of sentences as comment strings. No warning necessary then.

PR #924, opened 2024-03-18
update tensorqtl

zq2209:


gaow:
> parameter: maf_threshold = 0.05 if len(interaction) > 0 else MAC/(2.0*N) # if there is interaction, the maf can only be 0.05
@zq2209 why do we have to have this restriction? If you really want to limit it to 0.05 MAF you just put it in as a parameter?

gaow:
```
# Extract interaction name
if len(interaction) > 0:
    for item in interaction:
        if os.path.isfile(item):
            interaction_s = pd.read_csv(item, sep='\t', index_col=0)
            var_interaction = interaction_s.columns[0] # interaction name
        else:
            var_interaction = interaction
else:
    var_interaction = '' 
```

@zq2209 how many interaction terms are you allowing for? I think we only support one at a time? If that is the case, then `parameter:interaction = ""` should be an empty string by default?

gaow:
>         nominal = f'{cwd:a}/{_input[0]:bnn}{"_%s" % var_interaction if interaction else ""}.cis_qtl_nominal.txt.gz'

the added `txt` extesion does not seem to be necessary but am neutral about it. If you really want to add, better be `tsv` and should do this both for regional and for nominal to be consistent.

gaow:
```
    # Read interaction files or extract from covariates file. 
    for item in $[interaction]:
        if os.path.isfile(item):
            interaction_s = pd.read_csv(item, sep='\t', index_col=0)
            interaction_s = interaction_s[interaction_s.index.isin(covariates_df.index)] 
            var_interaction = interaction_s.columns[0] # interaction name
            # check if the interaction term in interaction table is in covariates file, if yes, extract it out from covariates file
            if var_interaction in covariates_df.columns:
                interaction_s[var_interaction] = covariates_df[var_interaction]
                covariates_df = covariates_df.drop(columns=[var_interaction])
            # drop samples that with missing value in iteraction
            interaction_s = interaction_s.dropna() 
        else:
            var_interaction = interaction
            interaction_s = covariates_df[var_interaction]
```

For the else part, don't you also drop interaction from covariates?

gaow:
As discussed, let's talk in person about this patch. Thanks!

gaow:
Regarding lambda_column for inflation factor: For when there is no interaction, lambda should be computed on nomial p-values. for when there is interaction, lambda should be computed for both main variant effect and for interaction effect, so there are one or two lambda columns depending on what analysis you do.

gaow:
Regarding permutatin: if it is False, what do we write for the output? Did you test it when you turn permuatoin equal False? Also as discussed if permutation does not support interaction, then we should skip permutation when doing interaction analysis. 

After you finish your coding, you need to make changes to various paramterers to test if things work; and under each paramter open the output and check if they make sense. For example if you allow interaction to be a list of values then your test should put in multiple variables (you might realize that does not make sense when you do it).

gaow:
For `cis_2`, it's okay to keep it. But let's remove the `bash` step of it that follows the python codes.

For the `try ... except` logic, do you think the Python code behave exactly the same as R? I don't think so. Unless you have tested and 100% sure that it is the correct code to put in, let's comment out the try .. except just put in the line that should work. We will analyze the data if it fails we will see exactly how it failed and to try .. except to catch that EXACT type of error, then address it. `except Exception as e` is too broad. You could have made a typo in your qvalue function call so it fail every time and you always use BH even though it is not about having too few p-vaues. 

`trans_2` can be removed altogether.

gaow:
The point to convert qvalue to Python is to add it to `cis_1` when permutation is not availabe in interaction analysis , or when we don't want to do permutations. You will need to use some sort of a group by approach to ocmpute qvalue for variants in the same region (ie gene) analyzed, and output an additional collumn of q-values. For interaction analysis you output extra two columns one for qvalue on the main effect one on interaction.

gaow:
@zq2209 looks good! Let me put some final touches on it then I think we are good to go. Thanks for fixing it up so quickly!

PR #930, opened 2024-03-20
gene annotation fixes for splicing

grennfp:


gaow:
@grennfp this patch seem to work but can be done better in terms of the codeing. I'll merge it for now so we can use. But we'll revisit it.

PR #943, opened 2024-03-26
fix tensorQTL bug

hsun3163:
In the previouse version, when using the genotype data from rosmap, there output for cis_2 will be a list of 22 chr genotype file connected by a space and produce an error. 

I don't believed this in intended so I fix that.


gaow:
@hsun3163 is it addressed here? https://github.com/cumc/xqtl-pipeline/pull/945 if so please close this PR. Thanks

PR #948, opened 2024-03-28
change the way of if condition for checking whether preset_variants_result has top_loci table for fsusie data 

kainblue100:
!is.null(dat_study[[condition]]$preset_variants_result$top_loci) is better

### DIFF
--- a/change_name.txt
+++ b/change_name.txt
@@ -1,1 +1,0 @@
-change origin branch name from xqtl-pipeline to xqtl-protocol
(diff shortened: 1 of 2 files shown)


change_name.txt
```diff
@@ -0,0 +1 @@
+change origin branch name from xqtl-pipeline to xqtl-protocol
```
gaow, line 1: @kainblue100 this file should not exist here right? Please remove.

gaow:
Closing this because we will not use preset results for this anyways in our downstreams analysis for fSuSiE. In the current version we removed this piece of code.

PR #961, opened 2024-04-09
update qvalue

zq2209:


gaow:
thanks @zq2209 ! a few comments:

1. `#from qvalue import qvalue` when it is something we surely don't use, don't feel bad to remove it. If you want this piece of code to be retrievable you can always give an informative commit message or to tag it so it lives in github history
2. `#pairs_df.columns.values[13]  = "pvalue" + '_' + var_interaction + '_' + 'interaction'` I prefer to make the change here, then remove this line: `names(pairs_df)[which(colnames(pairs_df) == "pval_gi")] = paste0('pvalue_', var_interaction, '_interaction')` you can achive this in SoS right? like `$["pvalue_%s_interaction" % var_interaction]` in place of `pval_gi` in your `R` codes.
3.  

```
  if ($['FALSE' if permutation and pval_threshold < 1.0 else 'TRUE']){ 
      pairs_df = pairs_df %>% group_by(molecular_trait_id) %>% mutate(qvalue = qvalue(pvalue)$qvalues)
```

too late to check it here. You should remove this if statement, but check it like:

```
stop_if(...)
R: expand= "$[ ]", stderr = f'{_output[0]}.stderr', stdout = f'{_output[0]}.stdout',container =container, entrypoint = entrypoint
...

```

we have used something like this logiic previoulsly right? Please learn from that and adopt.

PR #978, opened 2024-04-18
Change the output format of rss

SamuelSun2000:
1. File name change. 
2. add another layer of regions.
I've discussed with Ru and now we are satisfied with this format, also, pervious round result is already changed to this format manually


gaow:
thanks @SamuelSun2000 very helpful!

PR #1015, opened 2024-07-23
change gziped phenotype as bgziped format after sample lookup matching

rfeng2023:


gaow:
@rfeng2023 can you add at the start of the pipeline this, above the `input:` line:

```
depends: executable('tabix'), executable('bgzip')
```

PR #1027, opened 2024-08-16
quantile_twas pipeline + minor fix of mnm_regression and tensorqtl pipeline

al4225:
1. add quantile twas pipeline.
2. set symbolic link of quantile twas pipeline.
3. minor fix qvalue in tensorqtl: considering only 1 pvalue left.
4. minor fix def filter_by_region_ids in mnm_regression pipeline.


gaow:
Thanks @al4225 i merged it for now. My only comment is that maybe after things stabilize we can move the code to `pecotmr` package so you can document it better in roxygen format, provide unit test and vignettes as professional R packages do.

PR #1028, opened 2024-08-21
guard against strange snp name

hsun3163:
guard against the strange variant name,
also add back the parameter to filter x for the LD matrix computed on the fly


gaow:
@hsun3163 you have this line twice:

```
# compute LD from genotype
parameter: compute_LD_from_genotype = False
# remove a variant if it has more than imiss missing individual level data
parameter: imiss = 1.0
# MAF cutoff
parameter: maf = 0.0025
parameter: compute_LD_from_genotype = False
```

gaow:
@hsun3163 just a friendly reminder on this PR. 

Also for "guard against the strange variant name," can you use comment string to explain this a bit better in the source code?

PR #1030, opened 2024-08-22
Updated RNA_calling.ipynb

rl3328:
RNA_calling_interactive.ipynb is for interactive session; RNA_calling_mmjobman.ipynb is for mmjobman(submitted from local laptop). Versions for track changes.


rfeng2023:
image
image
image
@rl3328 I just found that my comments are all pending so you may not be able to see them. please see my comments in pictures.  Overall, I'd like you to keep a detailed record of which changes are necessary and which are temporary modifications to accommodate your environment.

rl3328:
1. I have let Frank know when I double-checked the filename with him.
2. I modified this part and the corresponding sos commands generator initializing mm_jobman.
3.4.6.7.8. Yes, these modifications are specific to interactive session. I also submitted a mm_jobman version that kept them unchanged.
5.9. I modified the hard-coded filenames to .Aligned.toTranscriptome.out{"_wasp_qc" if varVCFfile else ""}.bam
10.  I commented these functions out because they are not called in the whole pipeline.

rl3328:
The mmjobman version of RNA_calling might also need review for there are some modifications on output path. Thank you!

gaow:
As discussed @rl3328 please patch them directly to `RNA_calling.ipynb`. If we need to modify the image to use the latest environment, ie the enviroment you install / configure in the interactive session, we can work with Frank or Dan to update the container to reflect that.

code/molecular_phenotypes/calling/updated_RNA_calling/RNA_calling_interactive.ipynb
```diff
@@ -2523,7 +2523,7 @@
     "    #   return(metrics)\n",
     "    # }\n",
     "\n",
-    "\n",
+    "# I commented these functions out becuse they are not called in the whole pipeline.\n",
```
rfeng2023: I'd recommend to document your name and modified time

PR #1032, opened 2024-08-27
Changed on the original RNA_calling.ipynb to track changes

rl3328:
Changed on the original RNA_calling.ipynb to track changes and tried to unified two versions of code for interactive and mmjobman(but need more support on the inconsistencies due to environment differences)


gaow:
@rl3328 please remove the folder `updated_RNA_calling` to keep the patch neat. The only think i should focus on is your update to RNA_calling.ipynb

rl3328:
In summary, I changed:
1.  the read in order of sample_specific to correct the grouping of inputs
2. the output path of some files to make them organized and not permission denied
3.  some operations in the presence of varVCFfile, including the naming issues to make the pipeline applicable for both analysis with or without wasp_qc
4. some unused functions

gaow:
@rl3328 I suggest some changes in the interface, 
```
# list of samples
parameter: sample_list = path
# sample names to analyze
parameter: sample_name = list()
```

So you can specify either one or multiple samples.

Secondly, I still would like to have `wasp_filter()` as a separate step. Could you make it that way? 

>some unused functions

What do you mean by this? Could you elaborate a bit more?

Finally, it is the best you try your best to improve documentation on the page to reflect both changes you made, and anything you think is worth emphasizing. Imagine showing this to someone like yourself who is new to it and want to be able to follow through it just by reading the documentation.

rl3328:
Some unused functions include 1: some of the QC metrics that were defined but never called in this pipeline in [rsem_call_4, rnaseqc_call_4]. 2. the wasp_filter() function that you think simply commenting out makes the pipeline not neat and tidy.

rl3328:
I can add a [filter] step in the pipeline that filters tag[2:7].

PR #1040, opened 2024-09-05
minor fix

Chunmingl:


gaow:
this PR is not correct. We need to use --maf 0.01 --xvar 0.01 for `mnm`, not introducign another parameter. I will patch the x_variance separately.

PR #1059, opened 2024-09-16
Debugged the [cis_results_export] for fine_mapping_post_processing.ipynb

rl3328:
in previous command, condition within [['condition']] is recognized as a string instead of a variable (goover the conditions list), returning NULL.


code/post_processing/fine_mapping_post_processing.ipynb
```diff
@@ -1877,7 +1877,7 @@
     "          if(length(conditions) > 0) {\n",
     "  \n",
     "            for(condition in conditions) {\n",
-    "                dat_con <- dat_study${'' if multi_gene else [['condition']]}${[['fsusie_summary']] if fsusie else ''}\n",
+    "                if (${\"FALSE\" if multi_gene else \"TRUE\"}) dat_con <- dat_study[[condition]]${[['fsusie_summary']] if fsusie else ''} else dat_con <- dat_study${[['fsusie_summary']] if fsusie else ''}\n",
```
rfeng2023, line 1880: please try to change `[['condition']]` to `'[[condition]]'` or `[[condition]]` to make it looks cleaner. 

And if you want to keep this version, if you specify multi_gene, then it would not be fsusie mode, so the judgement in function in else{} can be deleted.

rl3328: Changed [['condition']] to '[[condition]]', added --step1_only in sos command and successfully tested 2 genes. But floating instance issue exists.

PR #1066, opened 2024-09-19
merge pipelines and clean up

Chunmingl:


gaow:
@Chunmingl this file `code/pecotmr_integration/python.ipynb` is it necessary?

Chunmingl:
This file is removed

gaow:
okay for ctwas the codes are also all commented out. I think at this point we should try to test the cTWAS codes and see if it works? Also does ctwas really depends on `sos_variable(regional_data)` because i have not seen it used?

Chunmingl:
the sos variables were updated after testing on twas and ctwas pipelines together

PR #1069, opened 2024-09-23
build quantas container

rfeng2023:
Hi @danielnachun I am trying to build quantas container to streamline analysis, for now I just need to add quantas and samtools into it, please let me know if there is any problem! Thanks!


danielnachun:
Looks good to me!

PR #1081, opened 2024-10-03
add vignette

grennfp:


gaow:
as discussed, let's move it outside of `MASH` folder and remove `mash_pipeline.ipynb` @grennfp

PR #1097, opened 2024-10-26
add region-name as an option in twas.ipynb

al4225:


gaow:
looks great! I appreciate it that you added documentations. It's good habit.

Down the road, as we improve our "export" function which will generate the meta-data table, we will consider xQTL type to include the information, so we can remove the needs for having that.

PR #1133, opened 2025-02-25
fix mnm name issue

rfeng2023:


gaow:
Oops indeed, I forgot about changing it down here. There may be a more elegant solution eg using "named output" feature in SoS but this would work. I think we also need to do this for --save-data and --fine-mapping outputs? When we add --save-data i think it is going to mess things up again.

rfeng2023:
> Oops indeed, I forgot about changing it down here. There may be a more elegant solution eg using "named output" feature in SoS but this would work. I think we also need to do this for --save-data and --fine-mapping outputs? When we add --save-data i think it is going to mess things up again?

indeed, will change all of them

PR #1137, opened 2025-03-04
add a step performing comprehensive finemapping(WIP)

rl3328:
1. It processes multiple GWAS datasets for specified genomic regions using a comprehensive fine-mapping strategy:
* Starts with SuSiE (Sum of Single Effects) RSS method
* Can switch to Bayesian Conditional Regression if two credible sets are found with low correlation(cs_corr<0.1)
* Can switch to Single Effect Regression if credible set is not found in SuSiE RSS but found in Single Effect Regression with GWAS marginal p_value for top variant(the variant in the cs that has the highest PIP) < 10^-5.
2. In `rss_analysis_pipeline()` in [pecotmr package](https://github.com/StatFunGen/pecotmr/blob/main/R/univariate_pipeline.R), `pip_cutoff_to_skip` process is executed before QC and imputation, I am wondering if I still need to move any code in pecotmr.


gaow:
This step looks like a lot duplicate with existing. We need to unify them (later).

PR #1158, opened 2025-03-25
fsusie interface update

hsun3163:


gaow:
Thanks @hsun3163 some comments:

> # Run fsusie with this post processing option , avaible option are TI, HMM, and none. TI is recommended as default. HMM perform particularly well when analysing data with say few sampling points (i.e ncol(Y)<30) or when the data are particularly noisy (low singla noise ratio).

typo "analyzing" and "signal"

> `cor_sample = False`

can we use `small_sample_correction = False` as input parameter to `cor_sample = ... ` for fSuSiE interface which is not a particualy good name

PR #1181, opened 2025-04-12
draft for comprehensive rss(WIP)

rl3328:
1. Generate the diagnostic table at the **CS** level, which provides the most detailed information.  
2. Add results from other methods based on the decisions guided by the diagnostic table.  
*Note: I’ve renamed `$RSS_QC_RAISS_imputed` according to the corresponding method.*


gaow:
thanks @rl3328 . After we talk more about this, we need to move the entire thing to `pecotmr`. There is a tiny detail regarding SER I also need to clarify with you. But let's first discuss the remaining issues on slack .

PR #1187, opened 2025-04-24
Create a pipeline to export finemapping results with automated QC(WIP)

rl3328:
1. I added the `auto_decision` step to the export pipeline.
2. I reordered parts of the code and added the conditional_effect column after the process_top_loci step. This change addresses the issue introduced by automated QC, where some blocks have top_loci but no credible sets (CS) across all methods or coverage levels.
3. I modified some documentations in mnm_postprocessing.ipynb.(container, pipeline name, etc.)
4. I added  a column to indicate the finemap_model used in the final exported top loci table.
5. Fixed a typo when I exporting Aging finemapping results.
To do:

6. To output diagnostic table for ColocBoost iteration. 
7. To integrate this updated pipeline into mnm_postprocessing.ipynb

I might fix auto_decision() function in pecotmr for susie_rss, since there's NA in z and p_value for the variants in diagnostics for PD, which is multi_ancestry. There might be missing variants in LD panel. Not sure if this is the reason. Will look into it further...


gaow:
@rl3328 as per our discussion yesterday, am closing this ticket because we are going to use the new approach from @SamuelSun2000

PR #1199, opened 2025-05-18
update sldsc pipeline: get heritability and meta analysis

pengjiao:


pengjiao:
1. Replaced the previous .M file generation code.
2. Added minor edits to include heritability and meta-analysis.

code/enrichment/sldsc_enrichment.ipynb
```diff
@@ -583,6 +583,8 @@
     "parameter: baseline_ld_dir = path()  # Directory containing baseline LD score files\n",
     "parameter: frqfile_dir = path()  # Directory containing allele frequency files\n",
     "parameter: weights_dir = path()  # Directory containing LD weights\n",
+    "parameter: baseline_name = \"baselineLD.\"  # Prefix of baseline annotation files\n",
+    "parameter: weight_name = \"weights.hm3_noMHC.\"  # Prefix of LD weights files\n",
```
gaow, line 587: @pengjiao `hm3_` stands for hapmap3 which we should not really include here. I wonder if we want to use a different weight name eg `sldsc_weights.noMHC.`

Speaking of MHC --- that reminds me, that our preprocessing should remove variants within hte MHC region on chr6?

code/enrichment/sldsc_enrichment.ipynb
```diff
@@ -992,8 +1012,10 @@
     "            cols <- paste0(\"ANNOT_\", annot_index) \n",
     "            num <- num + ((nrow(dat)-1) * sapply(dat[, ..cols], var))\n",
     "        } else {\n",
-    "            if((4 + annot_index) > ncol(dat)) stop(paste(\"Index out of range:\", m))\n",
-    "            num <- num + ((nrow(dat)-1) * var(dat[, 4+annot_index, with=FALSE]))\n",
+    "            # Newly changed: if((6 + annot_index) > ncol(dat)) stop(paste(\"Index out of range:\", m))\n",
```
gaow, line 1015: it is okay to remove all the `#Newly changed` hints since I can read the patch for it here to keep track. Please remove them.

code/enrichment/sldsc_enrichment.ipynb
pengjiao: I have two versions of weight files: weights.hm3_noMHC. (based on the 1000G EUR panel) and weight_chr (from the ADSP panel). Can I just rename weights.hm3_noMHC. to sldsc_weights.noMHC. for consistency?

